PR #3, opened 2020-01-08
Introduce generating configuration files from templates feature

cosmo0920:
Currently, the following files are installed into RPM:

* /etc/td-agent
* /usr/sbin


cosmo0920:
Generating configuration files are also based on treasure-data/omnibus-td-agent@9d86d39.

PR #6, opened 2020-01-09
Install unitfile and package scripts deb

cosmo0920:
* Generate package scripts for deb package
* Install systemd unit file and etc/td-agent/* files into deb package

TODO: gracefully td-agent service stop in deb package scripts


cosmo0920:
I've removed the generated deb package script files from this PR.

PR #26, opened 2020-04-09
Revert "RPM: Add missing dependencies"

ashie:
Reverts clear-code/td-agent-builder#25

It's my misuderstanding,  dependencies are automatically resolved.
I just forgot to enable EPEL:

$ sudo dnf install epel-release -y


cosmo0920:
Then, we should write down about EPEL dependency on CentOS 6/7/8.

PR #28, opened 2020-04-10
Add patches to remove old executable of racc in Ruby 2.7

ashie:
They take along with unexpected dependency since they contain wrong
shebang. In addition, they are already removed at upstream.

See also:
* https://bugs.ruby-lang.org/issues/15982
* https://github.com/ruby/racc/pull/123
* https://src.fedoraproject.org/rpms/ruby/c/baf046a6a4d17fa309c9d20fa3db949f6c24aacf?branch=master
* https://github.com/clear-code/td-agent-builder/pull/24


ashie:
The patches are backported from https://github.com/ruby/ruby

cosmo0920:
The built package doesn't contain `/usr/local` string:

```
% strings td-agent/yum/repositories/centos/8/x86_64/Packages/td-agent-3.7.0-1.el8.x86_64.rpm | grep usr/local 
```

PR #30, opened 2020-04-14
Bundle jemalloc

### LINKED ISSUE #29: Bundle jemalloc
https://github.com/clear-code/td-agent-builder/pull/24#issuecomment-612788868

> @ashie How about including jemalloc in td-agent?
> jemalloc changes internal memory management, so using same version is better for trouble-shooting.

ashie:
This commit discards "--with-jemalloc" option of Ruby, it seems hard to
use with libjemalloc in non-standard directory. Revive LD_PRELOAD
instead.

Fix #29

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -70,12 +70,14 @@ end
 class DownloadTask
   include Rake::DSL
 
+  attr_reader :file_jemalloc_source
   attr_reader :file_ruby_source, :file_ruby_installer_x64
   attr_reader :file_fluentd_archive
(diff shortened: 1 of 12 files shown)


ashie:
> This commit discards "--with-jemalloc" option of Ruby, it seems hard to use with libjemalloc in non-standard directory.

I'm also trying "--with-jemalloc" option in https://github.com/clear-code/td-agent-builder/commit/c04dae8eec4353fc887d5040f16046908de3e53b but not yet succeed. I met various issues with this option.

PR #33, opened 2020-04-16
yum: Remove EPEL dependecies

cosmo0920:
zstd and lz4 are provided from EPEL.

They should be removed from building environment and dependencies.

Related to https://github.com/clear-code/td-agent-builder/pull/24#issuecomment-614111428.


ashie:
Thanks!

Issue #31, opened 2020-04-15
Missing dependencies


ashie:
I compared dependency of librdkafka in both omnibus-td-agent and td-agent-builder.
I found td-agent-builder version lacks some dependencies:

omnibus-td-agent:

```
$ ldd ./embedded/lib/ruby/gems/2.4.0/gems/rdkafka-0.7.0/ext/librdkafka.so:
        linux-vdso.so.1 =>  (0x00007ffbffffe000)
        libm.so.6 => /lib64/libm.so.6 (0x00007efbf77ae000)
        libsasl2.so.3 => /lib64/libsasl2.so.3 (0x00007efbf7591000)
        libssl.so.1.0.0 => /opt/td-agent/embedded/lib/libssl.so.1.0.0 (0x00007efbf7325000)
        libcrypto.so.1.0.0 => /opt/td-agent/embedded/lib/libcrypto.so.1.0.0 (0x00007efbf6eec000)
        libz.so.1 => /opt/td-agent/embedded/lib/libz.so.1 (0x00007efbf6cd5000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007efbf6ad1000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007efbf68b5000)
        librt.so.1 => /lib64/librt.so.1 (0x00007efbf66ad000)
        libc.so.6 => /lib64/libc.so.6 (0x00007efbf62df000)
        /lib64/ld-linux-x86-64.so.2 (0x00007efbf7ddd000)
        libresolv.so.2 => /lib64/libresolv.so.2 (0x00007efbf60c6000)
        libcrypt.so.1 => /lib64/libcrypt.so.1 (0x00007efbf5e8f000)
        libgssapi_krb5.so.2 => /lib64/libgssapi_krb5.so.2 (0x00007efbf5c42000)
        libkrb5.so.3 => /lib64/libkrb5.so.3 (0x00007efbf5959000)
        libk5crypto.so.3 => /lib64/libk5crypto.so.3 (0x00007efbf5726000)
        libcom_err.so.2 => /lib64/libcom_err.so.2 (0x00007efbf5522000)
        libkrb5support.so.0 => /lib64/libkrb5support.so.0 (0x00007efbf5312000)
        libfreebl3.so => /lib64/libfreebl3.so (0x00007efbf510f000)
        libkeyutils.so.1 => /lib64/libkeyutils.so.1 (0x00007efbf4f0b000)
[...]

PR #35, opened 2020-04-16
yum: Add more dpendent libraries

cosmo0920:
Related to #31.

* libedit
* libtinfo
* libyaml

libxml2 and liblzma shouldn't be needed. Because we don't bundle nokogiri gem.


cosmo0920:
Thanks!

ashie:
liblzma and libxml aren't added because we're going to remove nokogiri, right?

cosmo0920:
Yes. Currently, they are not needed.

PR #36, opened 2020-04-16
apt: Add more dependent libraries

cosmo0920:
* libedit2
* libtinfo
* libyaml

I've added more dependent libraries for apt platforms.


cosmo0920:
Thanks!!

PR #39, opened 2020-05-07
deb: specify quilt format explicitly

kenhys:
Without this change, the format is regarded as native
package.

This change is followed DebSrc3.0 guideline.
ref. https://wiki.debian.org/Projects/DebSrc3.0


ashie:
Thanks!

PR #40, opened 2020-05-07
deb: add missing extended-description

kenhys:
It fixes E: td-agent: extended-description-is-empty
lintian error.


ashie:
Thanks!

PR #41, opened 2020-05-07
deb: set correct distribution in changelog

kenhys:
It fixes E: td-agent changes: bad-distribution-in-changes-file
unstable lintian error.

### DIFF
--- a/lib/apt/build.sh
+++ b/lib/apt/build.sh
@@ -75,7 +75,7 @@ run mkdir -p ~/.lintian/profiles/td-agent/
 run cp "debian/lintian/td-agent/${distribution}.profile" ~/.lintian/profiles/td-agent/${distribution}.profile
 # export DEB_BUILD_OPTIONS=noopt
 DISTRIBUTION=`lsb_release -c | cut -f2`
-sed -i'' -e "s/unstable/$DISTRIBUTION/" debian/changelog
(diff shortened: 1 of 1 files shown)


lib/apt/build.sh
```diff
@@ -74,6 +74,8 @@ fi
 run mkdir -p ~/.lintian/profiles/td-agent/
 run cp "debian/lintian/td-agent/${distribution}.profile" ~/.lintian/profiles/td-agent/${distribution}.profile
 # export DEB_BUILD_OPTIONS=noopt
+DISTRIBUTION=`lsb_release -c | cut -f2`
+sed -i'' -e "s/unstable/$DISTRIBUTION/" debian/changelog
```
ashie, line 78: This regular expression is too rough, all `unstable` string will be replaced.

kenhys: How about `sed -i'' -e "1s/ unstable;/ $DISTRIBUTION;/" debian/changelog` ?

ashie: It's still rough, should be matched more strictly. And older version should be also replaced instead of replacing only latest version.
Like this?:

`sed -E "s/^($PACKAGE \(\S+\)) unstable;/\1 $DISTRIBUTION;/g" changelog`

kenhys: Thanks, It is better to restrict expression like above.

kenhys: Fixed.

ashie:
Thanks!

PR #43, opened 2020-05-07
deb: drop needless Build-Depends

kenhys:
cdbs is not used in d/rules at all.
It fixes W: td-agent source: unused-build-dependency-on-cdbs

NOTE: cdbs is referenced from autotools indirectly,
(td-agent itself does not depends it directly)
So, cdbs is only still listed in Dockerfile.

### DIFF
--- a/td-agent/apt/debian-buster/Dockerfile
+++ b/td-agent/apt/debian-buster/Dockerfile
@@ -35,7 +35,6 @@ RUN \
   apt install -y -V ${quiet} \
     build-essential \
     debhelper \
-    cdbs \
     devscripts \
     ruby-dev \
     ruby-bundler \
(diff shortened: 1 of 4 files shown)


ashie:
> NOTE: cdbs is referenced from autotools indirectly,

Why autotools refer cdbs?
If you have any evidence for it, could you please show it?
For example it may be an error log with removing it from Dockerfile.

kenhys:
As build log is not left, I retried it without cdbs in Dockerfile again, then no build-depends error isn't occurred. Hmm, It is strange...
It is ok to remove cdbs from Dockerfile.

ashie:
> It is ok to remove cdbs from Dockerfile.

If there is no error or warning, please remove it from Dockerfile too.

kenhys:
> If there is no error or warning, please remove it from Dockerfile too.

It's ok so I've removed too.

ashie:
Thanks!

PR #47, opened 2020-05-11
deb: fix td-agent tmp files handling in postinst

kenhys:
It is false posivive because of /tmp/td-agent is
accessed by only td-agent:td-agent user/group.


ashie:
Could you fix a typo in the commit message?

```diff
-posivive
+positive
```

kenhys:
@ashie Thanks, I've fixed push -f.

ashie:
Thanks!

PR #48, opened 2020-05-11
deb: don't use recursive chown

kenhys:
It fixes W: maintainer-script-should-not-use-recursive-chown-or-chmod.
In postinst script, permission should be changed explicitly
to mitigate hardlink attacks.

### DIFF
--- a/td-agent/debian/td-agent.lintian-overrides
+++ b/td-agent/debian/td-agent.lintian-overrides
@@ -14,3 +14,4 @@ td-agent: duplicate-font-file
 td-agent: jar-not-in-usr-share
 td-agent: incorrect-libdir-in-la-file
 td-agent: possibly-insecure-handling-of-tmp-files-in-maintainer-script
+td-agent: maintainer-script-should-not-use-recursive-chown-or-chmod


td-agent/templates/package-scripts/td-agent/deb/postinst
```diff
@@ -57,10 +57,10 @@ update_conffile() {
     # changing to the right user. Then, these directories were created with
     # root permission. The following lines fix that problem.
     if [ -d "/var/log/<%= project_name %>/buffer/" ]; then
-        chown -R <%= project_name %>:<%= project_name %> /var/log/<%= project_name %>/buffer/
+        chown <%= project_name %>:<%= project_name %> /var/log/<%= project_name %>/buffer/*
```
ashie, line 60: I'm not sure there are grand-child directories or not.
So I think just suppressing the warning is better in this case.

kenhys: Ah, I will change PR.

kenhys:
Fixed.

ashie:
Thanks!

PR #49, opened 2020-05-11
deb: fix package-has-unnecessary-activation-of-ldconfig-trigger

kenhys:
dh_makeshlibs kicks unnecessary ldconfig triggers, so
it is suppressed by override_dh_makeshlibs.


ashie:
Thanks!

PR #55, opened 2020-05-15
deb: suppress W: incorrect-path-for-interpreter

kenhys:
ref. https://www.debian.org/doc/debian-policy/ch-files.html#scripts
It should be fixed in jemalloc upstream.


ashie:
Thanks!

PR #58, opened 2020-05-22
deb: remove needless .la files

kenhys:
It fixes E: td-agent: incorrect-libdir-in-la-file

No need to override td-agent: incorrect-libdir-in-la-file
in td-agent.lintian-override anymore.

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -664,6 +664,7 @@ class BuildTask
     Dir.glob("#{gem_staging_dir}/gems/*").each do |gem_dir|
       if File.exist?("#{gem_dir}/ext")
         rm_f(Dir.glob("#{gem_dir}/ext/**/*.o"))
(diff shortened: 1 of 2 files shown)


td-agent/debian/rules
```diff
@@ -13,6 +13,7 @@ override_dh_auto_install:
 	rake build:deb_config TD_AGENT_STAGING_PATH="$(CURDIR)/debian/tmp"
 	rake build:all        TD_AGENT_STAGING_PATH="$(CURDIR)/debian/tmp"
 	dh_installman
+	find $(CURDIR) -name '*.la' -print0 | xargs -0 rm -f
```
ashie, line 16: I want to remove these files from RPM too, so it should be done in Rakefile.

kenhys: Thanks, I'll fix it.

kenhys:
I've fixed to remove it in Rakefile.

PR #63, opened 2020-05-25
deb: suppress postinst warnings

kenhys:
Since dpkg 1.15.7.2, conffile handling is improved,
but to keep compatibility, update_conffile is still needed as is.


ashie:
Thanks!

Issue #66, opened 2020-05-25
deb: add option to skip lintian check


kenhys:
# Problem

Specific APT_TARGET doesn't finish to build.
There is no workaround for specific container currently.

# How to reproduce

rake apt:build APT_TARGETS="ubuntu-focal-arm64"

# Expected

rake apt:build APT_TARGETS="ubuntu-focal-arm64" success.
(It is able to skip lintian process if container's lintian is problematic version.)

# Actual

rake apt:build process stalls.

PR #68, opened 2020-05-25
deb: support to skip lintian

kenhys:
There is a case that lintian check stalls on ubuntu focal
arm64. Until that issue is fixed, skip lintian check as
a workaround.

Closes: #66

### DIFF
--- a/lib/apt/build.sh
+++ b/lib/apt/build.sh
@@ -80,13 +80,13 @@ if [ "${DEBUG:-no}" = "yes" ]; then
   if [ "${LINTIAN:-yes}" = "yes" ]; then
     run debuild -us -uc --lintian-opts --profile td-agent/${distribution}
   else
-    run debuild -us -uc --no-lintian
+    run debuild --no-lintian -us -uc
   fi
 else
   if [ "${LINTIAN:-yes}" = "yes" ]; then
     run debuild -us -uc --lintian-opts --profile td-agent/${distribution} > /dev/null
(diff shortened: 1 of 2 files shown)


kenhys:
waiting to pass CI

lib/apt/build.sh
```diff
@@ -79,7 +79,11 @@ sed -i'' -E "s/^($PACKAGE \(\S+\)) unstable;/\1 $DISTRIBUTION;/g" debian/changel
 if [ "${DEBUG:-no}" = "yes" ]; then
   run debuild -us -uc --lintian-opts --profile td-agent/${distribution}
 else
-  run debuild -us -uc --lintian-opts --profile td-agent/${distribution} > /dev/null
+  if [ "${LINTIAN:-yes}" = "yes" ]; then
```
ashie, line 82: I think lintian check should be done by default, skip it only when a user disables it.

kenhys: "user disables it" means setting LINTIAN=no environment variable by manual,  am I right?

ashie: Right.

kenhys: Thanks, I got it.

kenhys:
Oops, --no-lintian is ignored. I'll check again.

ashie:
> Oops, --no-lintian is ignored. I'll check again.

I've fixed it.

PR #69, opened 2020-05-25
rpm: fix licence field

kenhys:
It should be ASL 2.0 instead of Apache-2.0

ref. https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Software_License_List


ashie:
Thanks!

PR #74, opened 2020-05-27
Support Amazon Linux 2

### LINKED ISSUE #72: Support Amazon Linux 2
AWS provides Amazon Linux 2 images for amd64/arm64v8: https://hub.docker.com/_/amazonlinux

The problem is Amazon Linux 2 is similar to CentOS 7 but not the same.

- Need to use `amazon-linux-extras` to install ruby 2.4/2.6. No `centos-release-scl`.
- No `/etc/redhat-release`.  Maybe, fixed distribution/distribution_version is enough
- Need to update td-agent.spec.in for amazon linux 2
- May hit other difference point...

ashie:
Fix #72

### DIFF
--- a/lib/yum/build.sh
+++ b/lib/yum/build.sh
@@ -33,11 +33,11 @@ rpmbuild_options=
 
 . /host/env.sh
 
-distribution=$(cut -d " " -f 1 /etc/redhat-release | tr "A-Z" "a-z")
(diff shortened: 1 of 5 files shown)


cosmo0920:
Adding default yum targets and github action workflow for Amazon Linux 2 is next task?

PR #76, opened 2020-06-01
rpm: remove needless hidden file

kenhys:
It fixes W: hidden-file-or-dir

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -677,11 +677,9 @@ class BuildTask
     Dir.glob("#{gem_staging_dir}/gems/*").each do |gem_dir|
       cd(gem_dir) do
         rm_rf(["test", "spec"])
-        remove_files("**/\.*", true)
         remove_files("**/gem.build_complete")
         remove_files("ext/**/a.out")
         remove_files("ext/**/*.{o,la,a}")
-        remove_files("ext/**/*.{c,h}")
       end
     end
   end
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -667,6 +667,7 @@ class BuildTask
         rm_f(Dir.glob("#{gem_dir}/ext/**/*.o"))
         rm_f(Dir.glob("#{gem_dir}/ext/**/*.la"))
         rm_f(Dir.glob("#{gem_dir}/ext/**/*.a"))
+        rm_f(Dir.glob("#{gem_dir}/ext/**/a.out"))
         rm_f(Dir.glob("#{gem_dir}/ext/**/.dirstamp"))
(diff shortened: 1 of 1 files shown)


td-agent/Rakefile
```diff
@@ -665,11 +665,28 @@ class BuildTask
...
+      rm_f(Dir.glob("#{gem_staging_dir}/**/\.gemtest"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/\.editorconfig"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/\.readygo"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/\.rspec"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/\.ruby-version"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/\.travis"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/\.yardopts"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/\.*.yml"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/*.h"))
+      rm_f(Dir.glob("#{gem_staging_dir}/**/*.c"))
```
ashie, line 689: Could you check & clean up this method (`remove_needless_files`)?
Because current code performs redundant `glob`.
For example, `gem_staging_dir` shouldn't be used here since it's in the `Dir.glob("#{gem_staging_dir}/gems/*").each do |gem_dir|` block, and probably most of these files exist only under `#{gem_dir}` in this context.

kenhys: Thanks, I'll fix them

kenhys: I've rebased and push -f to fix them.

ashie: I've committed more simplified code.

td-agent/Rakefile
```diff
@@ -658,18 +658,30 @@ class BuildTask
...
-        rm_f(Dir.glob("#{gem_dir}/ext/**/*.la"))
-        rm_f(Dir.glob("#{gem_dir}/ext/**/.dirstamp"))
-        rm_f(Dir.glob("#{gem_dir}/ext/**/.sitearchdir*.time"))
+      cd(gem_dir) do
+        rm_rf(["test", "spec"])
+        remove_files("**/\.*", true)
+        remove_files("**/gem.build_complete")
+        remove_files("ext/**/a.out")
+        remove_files("ext/**/*.{o,la,a}")
+        remove_files("ext/**/*.{c,h}")
```
ashie, line 683: I'm now considering whether we can remove all dotfiles and C source files or not.

kenhys:
changed to keep .* and source code.

PR #78, opened 2020-06-01
rpm: fix W: name-repeated-in-summary

kenhys:
It seems that ", called td-agent" is redundant.

td-agent.x86_64: W: name-repeated-in-summary C td-agent
The name of the package is repeated in its summary.  This is often redundant
information and looks silly in various programs' output.  Make the summary
brief and to the point without including redundant information in it.


ashie:
Thanks!

PR #80, opened 2020-06-03
rpm: fix E: incoherent-logrotate-file

kenhys:
It should be /etc/logrotate.d/PACKAGE-NAME.


td-agent/Rakefile
```diff
@@ -324,7 +324,7 @@ class BuildTask
         configs = [
           "etc/td-agent/td-agent.conf",
           "etc/td-agent/td-agent.conf.tmpl",
-          "etc/logrotate.d/td-agent.logrotate",
+          "etc/logrotate.d/td-agent",
```
ashie, line 327: The original td-agent package copies /etc/logrotate.d/td-agent.logrotate to /etc/logrotate.d/td-agent if it's not exist: https://github.com/treasure-data/omnibus-td-agent/blob/master/templates/package-scripts/td-agent/rpm/post#L45

So the patch is no problem if it doesn't replace the old one, but it seems that `noreplace` isn't specified in the spec file: https://github.com/clear-code/td-agent-builder/blob/d99d91f6c87fb9f6a3ba2de3fb0f9f45a1ad4ad9/td-agent/yum/td-agent.spec.in#L115

kenhys: @ashie Thanks,
I've changed to add `noreplace` for it.

ashie:
Thanks!

ashie:
Oops, CI fails

```
RPM build errors:
    File not found: /root/rpmbuild/BUILDROOT/td-agent-3.7.1-1.el6.x86_64/etc/logrotate.d/td-agent.logrotate
Failed ./build.sh
```

ashie:
It's fixed via #85

PR #84, opened 2020-06-03
rpm: fix missing restart during upgrade

kenhys:


kenhys:
I'll merge #83 reduce conflicts.

PR #87, opened 2020-06-05
rpm: add missing owned directory

kenhys:
/var/log/td-agent/buffer

  ref. https://github.com/treasure-data/omnibus-td-agent/blob/master/templates/package-scripts/td-agent/rpm/post#L54

/etc/td-agent/plugin

ref. https://github.com/treasure-data/omnibus-td-agent/blob/master/templates/package-scripts/td-agent/rpm/post#L24


cosmo0920:
memo: `/tmp/td-agent` should be handled with tempfiles.d on CentOS 7 and 8.
https://developers.redhat.com/blog/2016/09/20/managing-temporary-files-with-systemd-tmpfiles-on-rhel7/

PR #92, opened 2020-06-10
rpm: remove needless static library

kenhys:
It fixes:
  W: spurious-executable-perm /opt/td-agent/lib/libjemalloc.a
  W: spurious-executable-perm /opt/td-agent/lib/libjemalloc_pic.a


ashie:
Thanks!

Issue #98, opened 2020-06-11
"The specified module could not be found" when using windows_eventlog2 


flurreN:
Everything works when using the [example windows_eventlog](https://github.com/fluent/fluent-plugin-windows-eventlog#in_windows_eventlog) config but not when im using my own or the [example windows_eventlog2](https://github.com/fluent/fluent-plugin-windows-eventlog#in_windows_eventlog2) config.

plugin versions:
```
winevt_c' 0.8.1
fluent-plugin-windows-eventlog 0.7.0
```
Tried on Windows 10 and Windows Server 2019 

Errorlog:
```
C:/opt/td-agent/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require': 126: The specified module could not be found.   - C:/opt/td-agent/lib/ruby/gems/2.4.0/gems/winevt_c-0.8.1/lib/winevt/winevt.so (LoadError)
        from C:/opt/td-agent/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from C:/opt/td-agent/lib/ruby/gems/2.4.0/gems/winevt_c-0.8.1/lib/winevt.rb:4:in `rescue in <top (required)>'
        from C:/opt/td-agent/lib/ruby/gems/2.4.0/gems/winevt_c-0.8.1/lib/winevt.rb:1:in `<top (required)>'
        from C:/opt/td-agent/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:133:in `require'
        from C:/opt/td-agent/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:133:in `rescue in require'
        from C:/opt/td-agent/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:39:in `require'
        from C:/opt/td-agent/lib/ruby/gems/2.4.0/gems/fluent-plugin-windows-eventlog-0.7.0/lib/fluent/plugin/in_windows_eventlog2.rb:1:in `<top (required)>'
        from C:/opt/td-agent/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from C:/opt/td-agent/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from C:/opt/td-agent/lib/ruby/gems/2.4.0/gems/fluentd-1.10.2/lib/fluent/registry.rb:102:in `block in search'
[...]

PR #103, opened 2020-06-13
Find and copy libstdc++-6.dll from building system

cosmo0920:
This will be fixing #98.

And this patch is inspired by https://github.com/treasure-data/omnibus-td-agent/blob/05a1f1a43732383828f4cbec47e393773c14acea/config/software/ruby.rb#L296-L323.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ td-agent-builder is a new build system for [td-agent](http://docs.treasuredata.c
   * Any host OS that is supported by Docker
     * Debian buster or Ubuntu 18.04 are recommended
   * [Docker](https://docs.docker.com/install/)
-  * Ruby 2.2 or later
+  * Ruby 2.4 or later
   * Git
 
 ### For building Windows package (.msi)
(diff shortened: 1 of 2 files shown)


cosmo0920:
And after merging this PR, td-agent-builder requests to use Ruby 2.4 or later which is installed by RubyInstaller2 based installer on Windows. Should we add this description in README?

ashie:
I think it's not but for the first aid.
In the future I want to resolve it more properly:

* Remove C++ dependency from winevt_c
  * It's not using so many C++ code, can be replaced with pure C easily.
* Detect other dependent libraries automatically

cosmo0920:
>     * Detect other dependent libraries automatically if they exist

For MSYS2, we can use this gemspec metadata:
https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#-msys2-library-dependency
We should use this metadata if depending non-bundled dynamic libraries.

cosmo0920:
I decided that this PR should be merged but after works should replace this "first aid".

PR #104, opened 2020-06-15
rpm: use tmpfiles.d for /var/run/td-agent/

kenhys:
It fixes: E: dir-or-file-in-var-run.

On CentOS 6, tmpfiles.d is not avairable, so to keep these directory,
create special file under such a directory.

Signed-off-by: Kentaro Hayashi <kenhys@gmail.com>

### DIFF
--- a/td-agent/templates/usr/lib/tmpfiles.d/td-agent.conf
+++ b/td-agent/templates/usr/lib/tmpfiles.d/td-agent.conf
@@ -1,4 +1,5 @@
 d /tmp/<%= project_name %> 0755 <%= project_name %> <%= project_name %> - -
+d /var/run/<%= project_name %> 0755 <%= project_name %> <%= project_name %> - -
 
 # Exclude td-agent
 x /tmp/<%= project_name %>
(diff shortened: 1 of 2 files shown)


kenhys:
special file is created expectedly.

```
$ ls -la /var/run/td-agent/
total 8
drwxr-xr-x.  2 td-agent td-agent 4096 2020-06-15 01:07 .
drwxr-xr-x. 12 root     root     4096 2020-06-15 01:06 ..
prw-r--r--.  1 root     root        0 2020-06-15 01:07 .keep
```

td-agent/templates/usr/lib/tmpfiles.d/td-agent.conf
```diff
@@ -1,4 +1,5 @@
 d /tmp/<%= project_name %> 0755 <%= project_name %> <%= project_name %> - -
+d /var/run/<%= project_name %> 0755 <%= project_name %> <%= project_name %> - -
```
cosmo0920, line 2: Nit: /var/run should be mounted as tmpfs. Shouldn't we preserve /var/run contents for td-agent?

kenhys: /var/run/td-agent/ is used to store td-agent.pid only, so I thought it is ok.

what "contents" means?

cosmo0920: > what "contents" means?

It means some files are put there.

PR #109, opened 2020-06-16
Split workflows

cosmo0920:
Using if condition branches is a bit messy work flow.
Splitting linux workflow into apt and yum should be simple.

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -1,4 +1,4 @@
-name: Linux
+name: Apt based Linux
 on:
   push:
   pull_request:
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -1,0 +1,28 @@
+name: Linux
+on:
+  push:
+  pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/apt.yml
```diff
@@ -0,0 +1,28 @@
+name: Linux
```
ashie, line 1: Pls set appropriate name.

cosmo0920: I'd used Apt based Linux and Yum based Linux.

kenhys: Debian and derivatives?

kenhys: except for `name:`, LGTM.

cosmo0920:
I'll merge this PR after Amazon Linux 2 package is built.

cosmo0920:
Amazon Linux 2 rpm package is built successfully:
https://github.com/fluent-plugins-nursery/td-agent-builder/actions/runs/136689355

PR #113, opened 2020-06-16
Build additional arm64 packages

cosmo0920:
We should also build Amazon Linux 2 and Ubuntu Focal arm64 packages.


cosmo0920:
https://github.com/fluent-plugins-nursery/td-agent-builder/releases/tag/testing-uploading-artifacts3 amazonlinux-2 and ubuntu-focal arm64 packages should be built when tagged.

cosmo0920:
We can deploy ARM 64 packages except for CentOS 7 when tagged.

PR #116, opened 2020-06-23
Use latest fluent-plugin-windows-eventlog

kenhys:
It fixes the issue that rc version of nokogiri (1.11 rc2) which supports ruby 2.7.1
is not installed

ref. https://github.com/sparklemotion/nokogiri/issues/2029


cosmo0920:
Thanks!

flavorjones:
Nokogiri v1.10.10 has been released with precompiled support for Ruby 2.7 on Windows.

PR #117, opened 2020-06-23
Use WiX makable release candidate version

cosmo0920:
Wix using package version should be modified into buildable version number.
This is one of the possible prerelease version implementation.

Signed-off-by: Hiroshi Hatake <cosmo0920.oucc@gmail.com>

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -244,6 +244,15 @@ class BuildTask
     @download_task = download_task
   end
 
+  def make_wix_version_numer(version)
+    if version.split("~")[1]
+      version_base = version.split("~")[0].chars.delete_if{|e| e == "."}.join("").to_i.pred.to_s.chars.join(".")
+      version_base + "." + Time.now.strftime("%m%d")
+    else
+      version
+    end
+  end
+
   def define
     namespace :build do
       desc "Install jemalloc"
@@ -396,7 +405,8 @@ class BuildTask
       task :wix_config do
         src  = File.join('msi', 'parameters.wxi.erb')
         dest = File.join('msi', 'parameters.wxi')
-        render_template(dest, src, template_params)
+        params = {wix_package_version: make_wix_version_numer(PACKAGE_VERSION)}
+        render_template(dest, src, template_params(params))
       end
 
       desc "Create configuration files for Red Hat like systems with systemd"
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -244,13 +244,12 @@ class BuildTask
     @download_task = download_task
   end
 
-  def make_wix_version_numer(version)
-    if version.split("~")[1]
-      version_base = version.split("~")[0].chars.delete_if{|e| e == "."}.join("").to_i.pred.to_s.chars.join(".")
-      version_base + "." + Time.now.strftime("%m%d")
-    else
-      version
-    end
+  def make_wix_version_number(version)
+    return version unless version.include?("~")
+    "%s.%s" % [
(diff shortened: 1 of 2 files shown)


td-agent/Rakefile
```diff
@@ -244,6 +244,15 @@ class BuildTask
     @download_task = download_task
   end
 
+  def make_wix_version_numer(version)
+    if version.split("~")[1]
```
kenhys, line 248: It may be better to restrict the number of splitting explicitly.

I mean `version.split("~", 2)`

td-agent/Rakefile
```diff
@@ -244,6 +244,15 @@ class BuildTask
...
+      version_base = version.split("~")[0].chars.delete_if{|e| e == "."}.join("").to_i.pred.to_s.chars.join(".")
```
kenhys, line 249: It seems that `delete(".")` is enough instead of `delete_if{...}`.

td-agent/Rakefile
```diff
@@ -244,6 +244,15 @@ class BuildTask
...
+      version_base + "." + Time.now.strftime("%m%d")
```
kenhys, line 250: What about:

```
def make_wix_version_number(version)
  return version unless version.include?("~")
  "%s.%s" % [
    version.split("~", 2)[0].delete(".").to_i.pred.to_s.chars.join("."),
    Time.now.strftime("%y%m%d")
  ]
end
```

td-agent/msi/parameters.wxi.erb
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Include>
-  <?define VersionNumber="<%= version %>" ?>
-  <?define DisplayVersionNumber="<%= version %>" ?>
+  <?define VersionNumber="<%= wix_package_version %>" ?>
+  <?define DisplayVersionNumber="<%= wix_package_version %>" ?>
```
kenhys, line 4: There is no need to unify to wix_package_version for DisplayVesionNumber.
I mean "DisplayVersionNumber == version", VersionNumber ==  wix_package_version.

td-agent/Rakefile
kenhys: Oops, need to remove "%y".

```
def make_wix_version_number(version)
  return version unless version.include?("~")
  "%s.%s" % [
    version.split("~", 2)[0].delete(".").to_i.pred.to_s.chars.join("."),
    Time.now.strftime("%m%d")
  ]
end
```

kenhys:
By the way, do we need to assume rc and public release in the same day? or multiple releases in the same day) if so, `Time.now.strftime("%m%d")` may not be appropriate...

cosmo0920:
> By the way, do we need to assume rc and public release in the same day? thus multiple releases in the same day) if so, `Time.now.strftime("%m%d")` may not be appropriate...

If release candidate and release package in the same day(2020/06/28), `4.0.0~rc1` should be interpreted as `3.9.9.0628` for release candidate version and release version should be remove `~rc1`. Then,`4.0.0` is used for released package version number. Conflicts between prerelease version and rc version does not occur.

td-agent/Rakefile
cosmo0920: Oh! Good suggestion. :+1:

td-agent/Rakefile
cosmo0920: Yep, Including `%y` should cause overflow for patch level version.
WiX version should be in 0 ~ 65535.

td-agent/msi/parameters.wxi.erb
cosmo0920: :+1:

PR #118, opened 2020-06-25
Support ~alphaN, ~betaN, and ~rcN version

kenhys:
Additionally, build time (hour) is also considered
for CI.

   ~rc1: x.x.x.10009 if ~rc1 is built at 9am
 ~beta1: x.x.x.1009  if ~beta1 is built at 9am
~alpha1: x.x.x.109   if ~alpha1 is built at 9am

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -246,9 +246,20 @@ class BuildTask
 
   def make_wix_version_number(version)
     return version unless version.include?("~")
+    revision = ""
+    case version
+    when /~rc(\d+)/
+      revision = $1.to_i * 10000 + Time.now.hour
+    when /~beta(\d+)/
+      revision = $1.to_i * 1000 + Time.now.hour
+    when /~alpha(\d+)/
+      revision = $1.to_i * 100 + Time.now.hour
+    else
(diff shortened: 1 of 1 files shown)


kenhys:
It is ok for DisplayVersion and VersionNumber as expected.

image

td-agent/Rakefile
```diff
@@ -246,9 +246,20 @@ class BuildTask
...
   def make_wix_version_number(version)
     return version unless version.include?("~")
+    revision = ""
+    case version
+    when /~rc(\d+)/
+      revision = $1.to_i * 10000 + Time.now.hour
+    when /~beta(\d+)/
+      revision = $1.to_i * 1000 + Time.now.hour
+    when /~alpha(\d+)/
+      revision = $1.to_i * 100 + Time.now.hour
```
cosmo0920, line 256: revison should be from 0 to 65535.
Should we check version number overflow here?

kenhys: I got it. I'll fix it.

kenhys: pushed commit to fix it.

PR #119, opened 2020-06-25
MSI: Install td-agent4 in conventional directory

kenhys:


kenhys:
Installed to program files.

image

kenhys:
Without proper configuration, td-agent.bat use /etc/fluent/fluent.conf by default.

```
c:/Program Files/Treasure Data/TD-Agent4/lib/ruby/gems/2.7.0/gems/fluentd-1.11.0/lib/fluent/config.rb:31:in `initialize': No such file or directory @ rb_sysopen - /etc/fluent/fluent.conf (Errno::ENOENT)
```

ashie:
I don't really want to install it to space-mixed folder such as `Program Files`.
Because plugins or gems might not work with such path.

kenhys:
I may be wrong. td-agent.bat should be used in this case.

```
2020-06-26 00:53:02 +0000 [info]: adding match pattern="td.*.*" type="tdlog"
2020-06-26 00:53:03 +0000 [warn]: #0 [output_td] Use different plugin for secondary. Check the plugin works with primary like secondary_file primary="Fluent::Plugin::TreasureDataLogOutput" secondary="Fluent::Plugin::FileOutput"
2020-06-26 00:53:03 +0000 [info]: adding match pattern="debug.**" type="stdout"
2020-06-26 00:53:04 +0000 [info]: adding source type="forward"
2020-06-26 00:53:04 +0000 [info]: adding source type="http"
2020-06-26 00:53:05 +0000 [info]: adding source type="debug_agent"
2020-06-26 00:53:05 +0000 [info]: #0 starting fluentd worker pid=6580 ppid=6396 worker=0
2020-06-26 00:53:07 +0000 [info]: #0 [input_debug_agent] listening dRuby uri="druby://127.0.0.1:24230" object="Fluent::Engine" worker=0
2020-06-26 00:53:07 +0000 [info]: #0 [input_forward] listening port port=24224 bind="0.0.0.0"
2020-06-26 00:53:07 +0000 [info]: #0 fluentd worker is now running worker=0
```

ashie:
> I may be wrong. td-agent.bat should be used in this case.

Yes, the path of td-agent.conf is specified by td-agent.bat

```bat
set FLUENT_CONF=%TD_AGENT_TOPDIR%\etc\td-agent\td-agent.conf
```

kenhys:
It is valuable it is installable under Program Files, but it may be introduce some troubles.
I'll close this PR.

PR #125, opened 2020-07-03
apt: Add ddeb copying targets

cosmo0920:
At least, Ubuntu Bionic and Focal renames dbgsym debian packages' extension to ddeb from deb.
We should copy them into package pool.

Signed-off-by: Hiroshi Hatake <hatake@clear-code.com>


cosmo0920:
Thanks！

PR #130, opened 2020-07-10
MSI: do not remove dll.a for C extensions

kenhys:


kenhys:
```
$ find . -name '*.dll.a'
./td-agent/lib/libx64-msvcrt-ruby270.dll.a
```

It keeps dll.a.

PR #131, opened 2020-07-10
MSI: Use ENV['TD_AGENT_TOPDIR'] to specify install directory

kenhys:
On windows, /var/... may creates C:/var, so it should be
ENV['TD_AGENT_TOPDIR'] (typically c:/opt/td-agent) prepended.

### DIFF
--- a/td-agent/templates/opt/td-agent/share/td-agent.conf.tmpl
+++ b/td-agent/templates/opt/td-agent/share/td-agent.conf.tmpl
@@ -16,12 +16,20 @@
   auto_create_table
   <buffer>
     @type file
+<% if windows? %>
+    path "#{ENV['TD_AGENT_TOPDIR']}/var/log/<%= project_name %>/buffer/td"
+<% else %>
     path /var/log/<%= project_name %>/buffer/td
+<% end %>
   </buffer>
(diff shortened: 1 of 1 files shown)


kenhys:
After starting service, /var is created under C:/opt/td-agent/var.

image

kenhys:
On buster: no ENV['TD_AGENT_TOPDIR'].

```
% cat /etc/td-agent/td-agent.conf | head -n 30
####
## Output descriptions:
##

# Treasure Data (http://www.treasure-data.com/) provides cloud based data
# analytics platform, which easily stores and processes data from td-agent.
# FREE plan is also provided.
# @see http://docs.fluentd.org/articles/http-to-td
#
# This section matches events whose tag is td.DATABASE.TABLE
<match td.*.*>
  @type tdlog
  @id output_td
  apikey YOUR_API_KEY

  auto_create_table
  <buffer>
    @type file
    path /var/log/td-agent/buffer/td
  </buffer>

  <secondary>
    @type file
    path /var/log/td-agent/failed_records
  </secondary>
</match>

## match tag=debug.** and dump to console
<match debug.**>
  @type stdout

```

PR #132, opened 2020-07-10
MSI: support to enable delay start

kenhys:
In the previous version (4.0.0), we need start
Fluentd Windows Service(fluentdwinsvc) manually.

In this commit, use delay start which is supported in Fluentd 1.11.1.


kenhys:
Delay start is enabled.

image

cosmo0920:
I’d confirmed that delayed starting is enabled.
winfluentdsvc-delayed-start

PR #140, opened 2020-07-13
deb: suppress lintian errors

kenhys:
It is caused by installing /opt.


cosmo0920:
I'll merge this PR after CIs are finished.

PR #141, opened 2020-07-14
Add installation testcases for apt/yum based Linux platform

cosmo0920:
I've added installation testcases for apt and yum based Linux platforms.


.github/workflows/apt.yml
```diff
@@ -8,11 +8,28 @@ jobs:
...
+        label:
+          - Debian GNU/Linux Buster amd64
+          - Ubuntu Xenial amd64
+          - Ubuntu Bionic amd64
+          - Ubuntu Focal amd64
+        include:
+          - label: Debian GNU/Linux Buster amd64
+            rake-job: debian-buster
+            test-docker-image: debian:buster
+            test-script: td-agent/apt/test.sh
```
kenhys, line 20: It seems that it is better to use meaningful file name.

For example. install-test.sh and so on.

cosmo0920: Fixed in https://github.com/fluent-plugins-nursery/td-agent-builder/pull/141/commits/db688448a4b1e6ecfbefe3555754ee70f509ccd4.

td-agent/yum/test.sh
```diff
@@ -0,0 +1,41 @@
...
+# CentOS 7 system-release-cpe is:
+# cpe:/o:centos:centos:7
+# CentOS 6 system-release-cpe is:
+# cpe:/o:centos:linux:6
+# This means that column glitch exists.
+# So, we should remove before "o" character.
+
+distribution=$(cat /etc/system-release-cpe | awk '{print substr($0, index($1, "o"))}' | cut -d: -f2)
+version=$(cat /etc/system-release-cpe | awk '{print substr($0, index($1, "o"))}' | cut -d: -f4)
+
```
kenhys: Another solution:

`$(cat /etc/system-release-cpe | sed -e 's/.\+o:\([a-z]\+\?\):.\+/\1/')`

and

`$(cat /etc/system-release-cpe | sed -e 's/.\+:\([0-9]\+\)$/\1/')`

cosmo0920: I feel that using awk is easier for me.

kenhys: okay.

Issue #145, opened 2020-07-15
deb: Failed to re-intall td-agent 4.0.0 when an old config file exists


ashie:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/fluentd/zBr5slN3HQU/7K0LJTEpBgAJ

```
---INITIAL INSTALL OF TD-AGENT4---

ubuntu@ip-10-0-1-167:~$ curl -L https://toolbelt.treasuredata.com/sh/install-ubuntu-bionic-td-agent4.sh | sh  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   851  100   851    0     0   4052      0 --:--:-- --:--:-- --:--:--  4071
==============================
 td-agent Installation Script
==============================
This script requires superuser access to install apt packages.
You will be prompted for your password by sudo.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3157  100  3157    0     0  33585      0 --:--:-- --:--:-- --:--:-- 33585
OK
Hit:1 http://us-west-2.ec2.ports.ubuntu.com/ubuntu-ports bionic InRelease
Hit:2 http://us-west-2.ec2.ports.ubuntu.com/ubuntu-ports bionic-updates InRelease
Hit:3 http://us-west-2.ec2.ports.ubuntu.com/ubuntu-ports bionic-backports InRelease
Get:4 http://packages.treasuredata.com/4/ubuntu/bionic bionic InRelease [7692 B]
Hit:5 http://ports.ubuntu.com/ubuntu-ports bionic-security InRelease
Get:6 http://packages.treasuredata.com/4/ubuntu/bionic bionic/contrib arm64 Packages [682 B]
Fetched 8374 B in 0s (23.5 kB/s)
Reading package lists... Done
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following NEW packages will be installed:
  td-agent
0 upgraded, 1 newly installed, 0 to remove and 27 not upgraded.
Need to get 9843 kB of archives.
After this operation, 57.3 MB of additional disk space will be used.
Get:1 http://packages.treasuredata.com/4/ubuntu/bionic bionic/contrib arm64 td-agent arm64 4.0.0-1 [9843 kB]
Fetched 9843 kB in 1s (15.6 MB/s)
[...]

PR #147, opened 2020-07-16
deb: update tmpl path

kenhys:
Closes: #145

Since TD Agent 4.0.0, tmpl path was changed from
/etc/td-agent/td-agent.conf.tmpl.
It is a good manner not to put .tmpl under /etc, but
path of new .tmpl is not updated.
It caused a bug that re-install package fails because of
the following error:

  Updating conffile /etc/td-agent/td-agent.conf ...
  cp: cannot stat '/etc/td-agent/td-agent.conf.tmpl': No such file or directory
  dpkg: error processing package td-agent (--configure):
   installed td-agent package post-installation script subprocess returned error exit status 1
  Errors were encountered while processing:
   td-agent
  E: Sub-process /usr/bin/dpkg returned an error code (1)

### DIFF
--- a/td-agent/templates/package-scripts/td-agent/deb/postinst
+++ b/td-agent/templates/package-scripts/td-agent/deb/postinst
@@ -69,7 +69,7 @@ case "$1" in
         add_system_user
         add_directories
         fixperms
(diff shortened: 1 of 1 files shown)


kenhys:
@cosmo0920  I've verified by https://github.com/fluent-plugins-nursery/td-agent-builder/issues/145#issuecomment-659105503

cosmo0920:
I'd confirmed that this issue has been fixed.

Issue #152, opened 2020-08-03
Use appropriate Gemfile.lock for each build environment


kenhys:
# Probrem

Until bundler 2.2.0?, Automultiplatform is not work as expected, so we can not
use correct .lock file for each build environment

# Expected

Don't remove .lock file and use proper .lock file separately.

# Actual

Currently, we remove .lock file

# Additional information

This issue was fixed in upstream, but not released yet.
https://github.com/rubygems/bundler/pull/7580

PR #154, opened 2020-08-03
Use appropriate Gemfile on each environment

kenhys:
In the previous version, installed gem may not be guaranteed on
build time because bundler automultiplatform is not fully supported
yet. (We need bundler 2.2.0 not released yet)

This commit fixes this issue by using separated Gemfile for each
environment.

Closes: #152


kenhys:
I'll do it.

kenhys:
I've updated README.md

cosmo0920:
> Should we write down the new dependent gems installation processes?
> Top Gemfile does not exist. So, we should change descriptions for installation procedures in README.
> I'd fallen this pitfall. cry

I mean that when start to use separated Gemfile, users must specify gemfile location with `--gemfile` option by bundler.
Just `bundle install` execution does not work.

cosmo0920:
So, `bundle install --gemfile /path/to/gemfile` should be added in README.

Issue #150, opened 2020-07-30
Nested "opt\td-agent" folder exists in MSI package


ashie:
```
C:\opt\td-agent>dir C:\opt\td-agent\opt\td-agent\share
 ドライブ C のボリューム ラベルは Windows です
 ボリューム シリアル番号は 8E7B-B5D5 です

 C:\opt\td-agent\opt\td-agent\share のディレクトリ

2020/07/09  16:51    <DIR>          .
2020/07/09  16:51    <DIR>          ..
2020/07/09  16:41                27 td-agent-ruby.conf
2020/07/09  16:41             2,508 td-agent.conf.tmpl
               2 個のファイル               2,535 バイト
               2 個のディレクトリ  115,490,639,872 バイトの空き領域
```

kenhys:
both of file are needless.

PR #156, opened 2020-08-03
MSI: Remove needless template files

kenhys:
It fixes nested "opt\td-agent" folder.

Closes: #150


kenhys:
No opt directory.

image

ashie:
I don't want to generate these files on Windows (instead of excluding generated files from the package).

PR #162, opened 2020-08-04
MSI: set TD_AGENT_TOPDIR correctly

kenhys:
Use TD_AGENT_TOPDIR in td-agent.conf on Windows
which is generated from .tmpl file.

and logrotate.d is not needed, so it was removed

NOTE: Alter installed at once, TD_AGENT_TOPDIR
is not refrected immediately.

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -379,15 +379,18 @@ class BuildTask
       desc "Create td-agent configuration files from template"
       task :td_agent_config do
         configs = [
-          "etc/td-agent/td-agent.conf",
-          "etc/logrotate.d/td-agent",
+          "etc/td-agent/td-agent.conf"
         ]
         configs.concat([
-          "opt/td-agent/share/td-agent.conf.tmpl",
-          "opt/td-agent/share/td-agent-ruby.conf"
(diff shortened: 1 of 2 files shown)


cosmo0920:
I'm building msi package with this branch.

cosmo0920:
I'd confirmed that `TD_AGENT_TOPDIR` should be set up in system environment variable:
td-agent4-td-agent-topdir-env-var

daipom:
Note: We need the system environmental variable because the service is launched by [winsvc.rb of Fluentd](https://github.com/fluent/fluentd/blob/d5685ada81ac89a35a79965f1e94bbe5952a5d3a/lib/fluent/winsvc.rb#L43-L48).

https://github.com/fluent/fluent-package-builder/pull/533#issuecomment-1644968227

daipom:
`TD_AGENT_TOPDIR ` was correctly reflected after this fix, but it is not still reflected immediately after installation.
The changes of environmental variables are not reflected during the installation process.

* https://learn.microsoft.com/en-us/windows/win32/msi/environment-table?redirectedfrom=MSDN#remarks

PR #166, opened 2020-08-05
Update default apt targets list

cosmo0920:
Ubuntu 20.04 and 16.04 are also built by default.

Signed-off-by: Hiroshi Hatake <hatake@clear-code.com>


cosmo0920:
Thanks!

PR #172, opened 2020-08-08
Handle fluentd revision on gemfile

cosmo0920:
I'd noticed that Fluentd revision information does not handle correctly on Gemfile.
This PR should fix it.
And I also update Gemfile.lock for both of Windows and Linux.


cosmo0920:
This PR is needed to proceed td-agent 4.0.1 release for Windows.

Issue #167, opened 2020-08-06
CI: rpm, more install package test case


kenhys:
# Problem

No piuparts related test senario.

# Expected

Similar to #148 more test is needed.

contents of package should be tested via #163 

# Actual

Only install step is tested.

PR #173, opened 2020-08-11
rpm: execute install/remove/upgrade test

kenhys:
Closes: #167

### DIFF
--- a/td-agent/yum/install-test.sh
+++ b/td-agent/yum/install-test.sh
@@ -34,8 +34,40 @@ case ${distribution} in
     ;;
 esac
 
+echo "INSTALL TEST"
 repositories_dir=/fluentd/td-agent/yum/repositories
 ${DNF} install -y \
   ${repositories_dir}/${distribution}/${version}/x86_64/Packages/*.rpm
 
 td-agent --version
+
+echo "UNINSTALL TEST"
+${DNF} remove -y td-agent
+
+echo "UPGRADE TEST from v3"
+rpm --import https://packages.treasuredata.com/GPG-KEY-td-agent
+case ${distribution} in
+  amazon)
+      cat >/etc/yum.repos.d/td.repo <<'EOF';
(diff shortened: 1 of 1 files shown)


kenhys:
@ashie Could you review this PR, please?

td-agent/yum/install-test.sh
```diff
@@ -14,11 +14,13 @@ set -exu
 distribution=$(cat /etc/system-release-cpe | awk '{print substr($0, index($1, "o"))}' | cut -d: -f2)
 version=$(cat /etc/system-release-cpe | awk '{print substr($0, index($1, "o"))}' | cut -d: -f4)
 
+ENABLE_UPGRADE_TEST=1
 case ${distribution} in
   amazon)
     case ${version} in
       2)
         DNF=yum
+	ENABLE_UPGRADE_TEST=0
```
ashie, line 23: indent

td-agent/yum/install-test.sh
```diff
@@ -29,13 +31,48 @@ case ${distribution} in
         ;;
       *)
         DNF="dnf --enablerepo=PowerTools"
+	ENABLE_UPGRADE_TEST=0
```
ashie, line 34: indent

td-agent/yum/install-test.sh
ashie: Is it O.K. to skip upgrading test for Amazon Linux 2? Following code seems trying to do it for this platform...

td-agent/yum/install-test.sh
kenhys: @ashie Thanks, I've fixed it.

td-agent/yum/install-test.sh
kenhys: @ashie Thanks, I've fixed wrong baseurl, so I've enabled test again.

Issue #195, opened 2020-09-04
README: Add bundler to prerequisites


ashie:
It's required on host OS since #142 is merged.

PR #200, opened 2020-09-15
Add missing bundler and apt source package explanation

kenhys:
Closes: #195

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,9 @@ td-agent-builder is a new build system for [td-agent](http://docs.treasuredata.c
 * Remove `embedded` directory by omnibus
   * Use `/opt/td-agent/bin/fluent-cat` instead of `/opt/td-agent/embedded/bin/fluent-cat`
 * Update core components: ruby, jemalloc and more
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -193,3 +196,8 @@ In System Preferences > Security & Privacy > Privacy > Accessibility, you should
 % rake yum:build
 % rake msi:build
 ```
+
+### Note for bump up the package version
+
+It assumes that Gemfile is used with Bundler's multiplatform feature,
+so bundler must be 2.2.0.rc1 or later.
```
cosmo0920, line 203: If Gemfile should be subject here and avoid to use passive voice:

```diff
-is used
+works
```

is correct what you mean?

---

It assumes that Gemfile *works* with Bundler's multiplatform feature,
so bundler must be 2.2.0.rc1 or later?

kenhys: Ah, It should be. I'll fix it.

kenhys: I've rebased with master.

cosmo0920 (approved):

PR #216, opened 2020-09-24
CI: use meaningful action name

kenhys:
Without name:, uses: is used by default.


cosmo0920 (approved):

cosmo0920:
I'll merge this PR after CIs are finished.

PR #226, opened 2020-10-26
msi: show td-agent and fluentd version

kenhys:
Example:

  > .\td-agent.bat --version
  td-agent 4.0.1 fluentd 1.11.2 (2f80f6cf0a4aa44b9795cd6a66894a6b8fa70d06)

### DIFF
--- a/td-agent/msi/assets/td-agent.bat
+++ b/td-agent/msi/assets/td-agent.bat
@@ -7,7 +7,9 @@ set TD_AGENT_VERSION=%TD_AGENT_TOPDIR%\bin\td-agent-version.rb
 for %%p in (%*) do (
     if "%%p"=="--version" (
         ruby "%TD_AGENT_VERSION%"
-        exit 0
+        gogo last
     )
 )
 "%TD_AGENT_TOPDIR%\bin\fluentd" %*
+
+:last
--- a/td-agent/msi/assets/td-agent.bat
+++ b/td-agent/msi/assets/td-agent.bat
@@ -3,4 +3,11 @@ set TD_AGENT_TOPDIR=%~dp0\..
 set PATH="%~dp0";%PATH%
 set FLUENT_CONF=%TD_AGENT_TOPDIR%\etc\td-agent\td-agent.conf
 set FLUENT_PLUGIN=%TD_AGENT_TOPDIR%\etc\td-agent\plugin
+set TD_AGENT_VERSION=%TD_AGENT_TOPDIR%\bin\td-agent-version.rb
+for %%p in (%*) do (
+    if "%%p"=="--version" (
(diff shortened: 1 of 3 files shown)


kenhys:
verified like this:

image

cosmo0920:
Yep, I'd confirmed that `td-agent.bat --version` or `cd ¥path¥to¥td-agent¥bin; td-agent --version` shows td-agent version and using Fluentd version.
But I also confirmed that Td-Agent Command Prompt is forcibly exited when executing `td-agent --version` with its default working directory.

WDYT?

td-agent/msi/assets/td-agent.bat
```diff
@@ -3,4 +3,11 @@ set TD_AGENT_TOPDIR=%~dp0\..
 set PATH="%~dp0";%PATH%
 set FLUENT_CONF=%TD_AGENT_TOPDIR%\etc\td-agent\td-agent.conf
 set FLUENT_PLUGIN=%TD_AGENT_TOPDIR%\etc\td-agent\plugin
+set TD_AGENT_VERSION=%TD_AGENT_TOPDIR%\bin\td-agent-version.rb
+for %%p in (%*) do (
+    if "%%p"=="--version" (
+        ruby "%TD_AGENT_VERSION%"
+        exit 0
```
cosmo0920, line 10: We should remove this line because exit 0 causes unexpected Td-Agent Command Prompt shutdown.

cosmo0920 (changes requested):

td-agent/msi/assets/td-agent.bat
cosmo0920: How about use `cmd /c "ruby %TD_AGENT_VERSION%"` here?

cosmo0920: Or, how about this?

```cmd
@echo off
set TD_AGENT_TOPDIR=%~dp0\..
set PATH="%~dp0";%PATH%
set FLUENT_CONF=%TD_AGENT_TOPDIR%\etc\td-agent\td-agent.conf
set FLUENT_PLUGIN=%TD_AGENT_TOPDIR%\etc\td-agent\plugin
set TD_AGENT_VERSION=%TD_AGENT_TOPDIR%\bin\td-agent-version.rb
for %%p in (%*) do (
    if "%%p"=="--version" (
        cmd /c "ruby %TD_AGENT_VERSION%"
        goto last
    )
)
"%TD_AGENT_TOPDIR%\bin\fluentd" %*

:last
```

kenhys: It looks good!

cosmo0920 (approved):
LGTM!

PR #230, opened 2020-11-04
rpm: use /run instead of /var/run for CentOS 7 or later

kenhys:


cosmo0920 (approved):
It's reasonable. :+1:
I'd also confirmed that this change can be buildable on my environment.

PR #233, opened 2020-12-04
Ci remove needless variables

ashie:


cosmo0920 (approved):
Looks good. Investigating for failing Windows CI tasks should be handled in another PR?

ashie:
Thanks!

> Investigating for failing Windows CI tasks should be handled in another PR?

I'm investigating it.

PR #236, opened 2020-12-14
Use stable bundler release 2.2.0

kenhys:
bundler 2.2.0  has been released.

ref. https://rubygems.org/gems/bundler/versions/2.2.0


kenhys:
Windows & Yum CI fails.

ashie:
> Windows & Yum CI fails.

They should be fixed in other pull requests.

PR #244, opened 2020-12-21
Fix an inappropriate error message

ashie:


ashie:
I'll fix it with CI error in another PR.

PR #247, opened 2020-12-25
windows: mitigate possible escalation of privileges

kenhys:
Closes: #3201

Reported by @zubrahzz

ref. https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-28169

In the previous version, NT AUTHORITY\Authenticated Users:(I)(M) is
granted. It means that logged in users can replace any files under
opt/td-agent/bin. It also allows for attacker to gain administrative
privileges by replacing these files because these files are executed
as a local services with SYSTEM privilege.

Note that this PR was merged, we need to update gems
by td-agent-gem with administraror privilege.

### DIFF
--- a/td-agent/msi/source.wxs
+++ b/td-agent/msi/source.wxs
@@ -27,6 +27,18 @@
       <Directory Id="WINDOWSVOLUME">
         <Directory Id="OPTLOCATION" Name="opt">
           <Directory Id="PROJECTLOCATION" Name="td-agent">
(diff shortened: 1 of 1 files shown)


kenhys:
The previleges are restricted:

image

cosmo0920:
> Note that this PR was merged, we need to update gems
> by td-agent-gem with administraror privilege.

I also confirmed that this concern and changed security attributes.

cosmo0920 (approved):

ashie (approved):

PR #250, opened 2021-01-26
Fix wrong cache path of bundler on installing gems

ashie:
Although the working directory on installing gems is
BUILDROOT\td-agent, the cache path is set to td-agent/download
so the resolved absolute path includes redundant "td-agent".
As a result pre-downloaded gems aren't used, they are downloaded
again unexpectedly.


kenhys (approved):
LGTM.

PR #255, opened 2021-02-02
Remove bundler's cache files that are Git origin

ashie:
For https://github.com/fluent-plugins-nursery/td-agent-builder/pull/248#issuecomment-771270276


ashie:
Something wrong. I'll retry.

PR #260, opened 2021-02-08
First aid to avoid installing mismatched architecture's gems

ashie:
When a source tarball is created by a host OS that is different from
a target, fat gems for the host OS are brought to the build container
and installed. To avoid it, remove gems from a tarball so that
appropriate gems are downloaded by the build container.


ashie:
Unified with #261

PR #261, opened 2021-02-09
Use bundler to install gems

ashie:
Previous code uses `gem install` to install gems so that it ignores gem's versions described in Gemfile.lock. Because we are now using Gemfile, there is no reason not to use bundler to install gems.
In addition, this PR also prevent to install mismatched architecture's gem even if a source taball is created by a host OS that is different from a target.

### DIFF
--- a/td-agent/config.rb
+++ b/td-agent/config.rb
@@ -9,6 +9,8 @@ JEMALLOC_VERSION = "5.2.1"
 # https://www.openssl.org/source/
 OPENSSL_VERSION = "1.1.1i"
 
+BUNDLER_VERSION= "2.2.0"
+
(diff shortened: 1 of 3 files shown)


td-agent/config.rb
```diff
@@ -9,6 +9,8 @@
 # https://www.openssl.org/source/
 OPENSSL_VERSION = "1.1.1i"
 
+BUNDLER_VERSION= "2.2.0"
+
```
kenhys, line 13: How about using the latest bundler 2.2.9?
It seems that performance regression is resolved up to 2.2.9.

ashie: I'll do it in another PR.

kenhys: I got it.

PR #262, opened 2021-02-09
Remove needless ri documents in RubyInstaller (as of 2.7.2-1)

ashie:
The size of windows installer is decreased from 35MB to 26MB.


kenhys (approved):
LGTM

PR #264, opened 2021-02-15
CI: remove unused Travis related gem

kenhys:
As CI had been migrated into GitHub Actions, it should be removed.


ashie:
Please remove also rake tasks that uses the gem.

ashie:
Probably the following TODO can be also removed:
https://github.com/fluent-plugins-nursery/td-agent-builder/blob/master/TODO.md#enable-lintian-check-on-ubuntu-focal-arm64

kenhys:
Oops, I've committed directly. orz...

kenhys:
I've rebased.

PR #265, opened 2021-02-15
Explicitly install webrick 1.7.0

kenhys:
Because of Ruby 3.0 dropped it.


kenhys:
When webrick is bundled, td-agent msi is correctly installed without rolling back.


```
PS C:\work\td-agent-builder\td-agent-builder.work> C:\opt\td-agent\bin\ruby.exe -v
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x64-mingw32]
PS C:\work\td-agent-builder\td-agent-builder.work> C:\opt\td-agent\bin\td-agent-gem.bat list |select-string webrick

webrick (1.7.0)
````

PR #267, opened 2021-02-15
Raise IPV4 priority for connecting rubygems.org

kenhys:
If IPv6 is preferred in container, we can not
connect to rubygems.org. it causes timeout error.

### DIFF
--- a/.github/workflows/apt-arm.yml
+++ b/.github/workflows/apt-arm.yml
@@ -28,7 +28,7 @@ jobs:
           sudo apt update
           sudo apt -V install ruby ruby-bundler ruby-serverspec
           sudo apt -V install qemu-user-static
-          sudo gem install bundler --no-document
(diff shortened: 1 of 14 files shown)


kenhys:
CI is broken, need to fix it.

kenhys:
This failure is caused by bundle config set.

```
  def build_include_binaries_file
    # Note: maintain debian/source/include-binaries manually is inappropriate way
    # because many gem files are bundled with td-agent package.
    # We use gem specification GEMFILE to solve this issue.
    paths = []
    command = "bundle config set --local cache_path #{DOWNLOADS_DIR}"
    output, error, status = Open3.capture3(command)
    unless status.success?
      fail "Failed to set cache_path: <#{command}> (stdout: #{output}, stderr: #{error})"
    end
```

kenhys:
installed version of bundler must be matched with Gemfilel.lock.

PR #268, opened 2021-02-16
Support Ubuntu 21.04 (Hirsute)

kenhys:
#267 should be merged beforehand.


kenhys:
I've rebased with latest master.

PR #271, opened 2021-02-25
RPM: Remove conflicted `provides` libraries

ashie:
The following libraies that provided by td-agent 4 conflict with system's one:

  * libjemalloc.so.2()(64bit)
  * librdkafka.so.1()(64bit)
  * libruby.so.2.7()(64bit)

All shared libraries under /opt/td-agent/ should be ignored.
Otherwise td-agent may be installed unexpectedly when a package that requires these libraries is installed.

### DIFF
--- a/td-agent/yum/td-agent.spec.in
+++ b/td-agent/yum/td-agent.spec.in
@@ -17,6 +17,9 @@
 # specific language governing permissions and limitations
 # under the License.
 
+%global __provides_exclude_from ^/opt/%{name}/.*\\.so.*
+%global __requires_exclude libjemalloc.*|libruby.*|/opt/%{name}/.*
+
 %define _centos_ver %{?centos_ver:%{centos_ver}}%{!?centos_ver:8}
 %define _amazon_ver %{?amzn:%{amzn}}%{!?amzn:0}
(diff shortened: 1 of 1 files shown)


kenhys (commented):
It seems that it is better to place `%global` macro in the head.
(treat like %define macro)

ashie:
Thanks, I moved it.

kenhys:
It seems that needless provides and requires are removed.

```
% rpm -q --provides ./td-agent/yum/repositories/centos/8/x86_64/Packages/td-agent-4.1.0-1.el8.x86_64.rpm             
config(td-agent) = 4.1.0-1.el8
td-agent = 4.1.0-1.el8
td-agent(x86-64) = 4.1.0-1.el8
```

```
% rpm -q --requires ./td-agent/yum/repositories/centos/8/x86_64/Packages/td-agent-4.1.0-1.el8.x86_64.rpm |\grep  libr
librt.so.1()(64bit)
librt.so.1(GLIBC_2.3.3)(64bit)
% rpm -q --requires ./td-agent/yum/repositories/centos/8/x86_64/Packages/td-agent-4.1.0-1.el8.x86_64.rpm |\grep  libj
```

PR #273, opened 2021-02-26
WIP: Retry bundle package when bundle command has failed

kenhys:
There is a case that bundle fails from time to time.


ashie:
Can your problem be fixed with #276?

kenhys:
I'm not sure yet, but it will be fixed.
I'll close this WIP.

PR #276, opened 2021-03-01
CI: Remove installed fileutils gem before downloading gems

ashie:
Avoid following error on build_include_binaries_file

> stderr: Could not find fileutils-1.5.0.gem for installation

TODO: Should succeed to download it even if system's one exists


ashie:
I confirmed that it can avoid this issue but I'm now considering more appropriate solution.

ashie:
Windows test will fail due to another reason (package size check will fail because 4.1.0 for Windows isn't released yet).

ashie:
Not the best but better than before. We should revive CI ASAP.

PR #279, opened 2021-03-03
centos-6: Disable obsolete mirrorlist for yum

ashie:


ashie:
CI isn't prepared for it because CentOS 6 is already EOL.
I confirmed that it can build the package on my local environment.

PR #280, opened 2021-03-03
fix ci for windows, use heroku uri to check instead

ksandrmatveyev:
to fix  https://github.com/fluent-plugins-nursery/td-agent-builder/issues/278

@ashie , fyi


ashie:
Windows CI passed, thanks for your contribution!

PR #289, opened 2021-04-09
msi: fix uninstallable binary on non-English

kenhys:
Reported via https://github.com/fluent/fluentd/issues/3322

For example, there is a case that the required account is not created by
non-english platforms. It depends on localization.

Instead, it may be better to use well-known ACEs in SDDL.

ref. https://docs.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-definition-language-for-conditional-aces-

### DIFF
--- a/td-agent/msi/source.wxs
+++ b/td-agent/msi/source.wxs
@@ -39,12 +39,28 @@
                     NT AUTHORITY\SYSTEM:(OI)(CI)F
 
                   ref. https://wixtoolset.org/documentation/manual/v3/xsd/wix/permissionex.html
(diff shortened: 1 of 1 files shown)


kenhys:
This PR is not tested.

kenhys:
According to https://github.com/fluent/fluentd/issues/3322#issuecomment-816532142, 
It works without a workaround, so this fix is reasonable.

ashie:
I've added more comments to make easy to understand the curse spells :imp:

ashie:
Thanks!

PR #293, opened 2021-04-19
Fix missing td command in package

kenhys:


ashie:
For linux, it should be done at build time instead of install time because the path is never changed.
On Windows, the install path may be changed by a user, this is why windows installer do it at install time.

kenhys:
Hmm, this approach should be changed.

kenhys:
Changed to use gem pristine correctly during the build process.

ashie:
I confirmed that td and elastic_ruby_console in a built package works correctly.

ashie:
Merged. Thanks!
BTW please consider adding a test for it :)

kenhys:
I'll add a test case in another PR.

PR #297, opened 2021-04-24
td-agent-apt-source: Refer component variable

cosmo0920:
This should be needed for Debian Buster td-agent-apt-source package.
Otherwise, apt complains the following component misspelt error:

xxx doesn't have the component 'contrib' (component misspelt in sources.list?)

Signed-off-by: Hiroshi Hatake <cosmo0920.oucc@gmail.com>

---

This bug can be found with a created repository by aptly on AWS S3.


cosmo0920:
This fix is not urgent but the Debian future release should be removed apt-key command, right?
If so, we should prepare to create repository import packages with "correct package".

cosmo0920:
Hmm..., td-agent repository does not use main component even if debian buster package... That's weird convensions....
I'm closing for now. :cry:

PR #302, opened 2021-04-27
apt: drop apt-key from test script

kenhys:
As apt-key will be removed in Debian 12, it is not recommended to use
apt-key. The replacement of apt-key is using /etc/apt/trusted.gpg.d/
keyring instead.

NOTE: GnuPG 2.x new keyring format (keybox) is not recognized, we need
to use legacy public ring v4.x for a moment.

See
https://www.debian.org/releases/testing/amd64/release-notes/ch-information.en.html#obsolescense-and-deprecation

### DIFF
--- a/td-agent/apt/serverspec-test.sh
+++ b/td-agent/apt/serverspec-test.sh
@@ -16,16 +16,18 @@ td-agent --version
 
 case ${code_name} in
     xenial)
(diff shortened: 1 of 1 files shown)


kenhys:
I'll tweak a bit.

kenhys:
It should be now fixed. waiting CI.

PR #308, opened 2021-06-23
Fix a bug that fluentd gets stuck on DNS lookup due to resolv's bug

ashie:
See following links for more detail:
* https://bugs.ruby-lang.org/issues/17748
* https://github.com/fluent/fluentd/issues/3382

The patch for Ruby is taken from the following commit:
https://github.com/ruby/ruby/commit/9edc162583a4f685332239f6249745ad9b518cbe

Signed-off-by: Takuro Ashie <ashie@clear-code.com>

PR #312, opened 2021-06-25
Apply the patch for resolv also on Windows

ashie:
Follow up for #308

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -389,6 +389,7 @@ class BuildTask
       task :ruby => [:jemalloc, :openssl, :"download:ruby"] do
         if windows?
           extract_ruby_installer
+          apply_ruby_installer_patches
           setup_windows_build_env
           find_and_put_dynamiclibs
         else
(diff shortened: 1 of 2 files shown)


kenhys:
ok, I'll check it.

td-agent/Rakefile
```diff
@@ -763,6 +764,25 @@ class BuildTask
     end
   end
 
+  def apply_ruby_installer_patches
+    return if BUNDLED_RUBY_INSTALLER_PATCHES.nil?
+
+    version = BUNDLED_RUBY_INSTALLER_X64_VERSION.gsub(/-\d+$/, '')
+    feature_version = BUNDLED_RUBY_INSTALLER_X64_VERSION.gsub(/\d+-\d+$/, '0')
+    ruby_lib_dir = File.join(td_agent_staging_dir, "lib", "ruby", feature_version)
```
kenhys, line 772:
```
version = BUNDLED_RUBY_INSTALLER_X64_VERSION.split(/-/, 2).first
feature_version = version.sub(/\d+$/, '0')
```

It may be better to apply sub operation to  a reduced target (version)

ashie: Thanks, I've fixed it.

ashie:
I confirmed that the patch is applied expectedly.

PR #316, opened 2021-07-12
Update Ruby versions

ashie:
3.0.1 -> 3.0.2
2.7.3 -> 2.7.4
2.6.7 -> 2.6.8


ashie:
I don't remove td-agent/patches/ruby-3.0/0001-ruby-resolv-Fix-confusion-of-received-response-messa.patch for now, because it's can become an example for patching RubyInstaller.

PR #317, opened 2021-07-13
Remove jeprof

### LINKED ISSUE #306: Remove unnecessary perl depencencies in rpm packages
Are perl dependencies really necessary for fluentd to work ?

When installing 4.1.1-1.el7 on rhel 7, it requires 27 perl packages that I think that may be unnecessary.

Checking yum/td-agent.spec.in history, I see that some work on that direction has been done for CentOS 6 in [5c90584](https://github.com/fluent-plugins-nursery/td-agent-builder/commit/5c905844edf9e4a2db556ef9ba16f4e3c4ea9c76). Could this be done for CentOS 7 too?

ashie:
It won't be used in production environments.
And it introduces unnecessary dependencies.

Fix #306

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -1045,6 +1045,7 @@ class BuildTask
   end
 
   def remove_needless_files
+    remove_files("#{td_agent_staging_dir}/bin/jeprof", true)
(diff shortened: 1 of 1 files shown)


cosmo0920 (approved):
LGTM

kenhys (approved):
LGTM

Issue #176, opened 2020-08-13
Failed to start td-agent with aarch64


rluisr:
Summary
========
Failed to start td-agent **4.0.0** on CentOS 8 and 7.

```
# cat /var/log/messages

Aug 13 18:18:00 ip-10-105-44-194 fluentd[15556]: <jemalloc>: Unsupported system page size
Aug 13 18:18:00 ip-10-105-44-194 fluentd[15556]: <jemalloc>: Unsupported system page size
Aug 13 18:18:00 ip-10-105-44-194 fluentd[15556]: [FATAL] failed to allocate memory
```

Reproduce
=========
```
# yum install http://packages.treasuredata.com.s3.amazonaws.com/4/redhat/8/aarch64/td-agent-4.0.0-1.el8.aarch64.rpm -y
# systemctl start td-agent
```

Environnment
===========
- CentOS Linux release 8.2.2004 (Core)
- aarch64
- AWS EC2 ARM instance (M6g)

Please let me know if any of this information is missing, as I may be missing some of it.

kenhys:
@rluisr it may be that jemalloc is not compiled appropriately on aarch64 (not sure yet).

As a workaround, 
just comment out `#Environment=LD_PRELOAD=/opt/td-agent/lib/libjemalloc.so` in `td-agent.servce` then execute systemctl daemon-reload && systemctl start td-agent works as expected. (even though jemalloc can't be enabled :sob: )

```
$ cat /lib/systemd/system/td-agent.service 
[Unit]
Description=td-agent: Fluentd based data collector for Treasure Data
Documentation=https://docs.treasuredata.com/articles/td-agent
After=network-online.target
Wants=network-online.target

[Service]
User=td-agent
Group=td-agent
LimitNOFILE=65536
#Environment=LD_PRELOAD=/opt/td-agent/lib/libjemalloc.so
Environment=GEM_HOME=/opt/td-agent/lib/ruby/gems/2.7.0/
Environment=GEM_PATH=/opt/td-agent/lib/ruby/gems/2.7.0/
Environment=FLUENT_CONF=/etc/td-agent/td-agent.conf
Environment=FLUENT_PLUGIN=/etc/td-agent/plugin
Environment=FLUENT_SOCKET=/var/run/td-agent/td-agent.sock
Environment=TD_AGENT_LOG_FILE=/var/log/td-agent/td-agent.log
Environment=TD_AGENT_OPTIONS=
[...]

PR #318, opened 2021-07-13
jemalloc: Downgrade to 3.6.0 to reduce memory usage

### LINKED ISSUE #305: Consider replacing (or adding) jemalloc with v3.6.0
From https://github.com/fluent/fluentd/issues/1657

Some users reported that jemalloc v4 or v5 consumes excessive amounts of memory and v3.6 quite reduces it. It seems a known issue as follows:

* https://bugs.ruby-lang.org/issues/14718
* https://www.joyfulbikeshedding.com/blog/2019-03-14-what-causes-ruby-memory-bloat.html
* https://www.speedshop.co/2017/12/04/malloc-doubles-ruby-memory.html

> Do not use jemalloc 4.x with Ruby. It has a bad interaction with Transparent Huge Pages that reduces the memory savings you’ll see. Instead, use jemalloc 3.6. 5.0’s performance with Ruby is currently unknown.

Replacing (or adding) jemalloc with v3.6.0 isn't hard, but I think there is a points to be considered.
Is v3.6 still supported and can we get security fixes (it seems stable enough though)?
It's also concerned in the above ruby-lang's issue:

https://bugs.ruby-lang.org/issues/14718#note-37
```
it will be harder to get support from the jemalloc team if we are not on 5.0.1. 
```

Enterprise users are sensitive in security issues so it would be better that we have evidence about it. According to jemlloc's web site, it still seems be supported:

http://jemalloc.net/
> stable-3/stable-4: The stable-* branches are bugfix-only branches based on the last releases for previous major versions. The versioned release history is linear, but relevant fixes may continue to be applied to the stable-* branches indefinitely.

ashie:
Fix #305

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -631,21 +631,25 @@ class BuildTask
     configure_opts = [
       "--prefix=#{install_prefix}",
     ]
-    if ENV["TD_AGENT_STAGING_PATH"] and
-      (ENV["TD_AGENT_STAGING_PATH"].end_with?("el8.aarch64") or
-       ENV["TD_AGENT_STAGING_PATH"].end_with?("el7.aarch64"))
-      # NOTE: There is a case that PAGE_SIZE detection on
-      # CentOS 7 CentOS 8 with aarch64 AWS ARM instance.
-      # So, explicitly set PAGE_SIZE by with-lg-page 16 (2^16 = 65536)
-      configure_opts.concat(["--with-lg-page=16"])
+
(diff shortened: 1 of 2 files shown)


ashie:
I confirmed that we don't need to care the page size issue like #176 when we use jemalloc 3.6.
Built binaries are works fine on Amazon EC2 aarch64 instances.

ashie:
I tested jemalloc 3.6 and I don't have any problem with this for now.
I decide to ship next version with this.

PR #320, opened 2021-07-20
convert-artifacts-layout.sh: Don't skip AmazonLinux AArch64

ashie:
It's also published.


kenhys (approved):
LGTM

PR #322, opened 2021-07-26
rpm yum: use os-release

kenhys:
os-release is machine readable and contains consistent identification
data.

ref. https://www.freedesktop.org/software/systemd/man/os-release.html

### DIFF
--- a/lib/yum/build.sh
+++ b/lib/yum/build.sh
@@ -33,13 +33,27 @@ rpmbuild_options=
 
 . /host/env.sh
 
-distribution=$(cut -d " " -f 1 /etc/system-release | tr "A-Z" "a-z")
-if grep -q Linux /etc/system-release; then
-  distribution_version=$(cut -d " " -f 4 /etc/system-release)
-elif grep -q Stream /etc/system-release; then
-  # ${version}-stream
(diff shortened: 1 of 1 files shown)


kenhys:
This PR is extracted from https://github.com/fluent-plugins-nursery/td-agent-builder/pull/321

kenhys:
It is considered with building for CentOS 6.

lib/yum/build.sh
```diff
@@ -33,13 +33,27 @@ rpmbuild_options=
...
+  case $ID in
+    amzn)
+      # The ID is translated for repository layout
+      distribution=amazon
+      ;;
+    *)
+      distribution=$ID
+      ;;
+  esac
+  if grep -q Stream /etc/os-release; then
```
cosmo0920, line 47: Using grep is also OK.
Why don't you use `$NAME` for detecting `CentOS Stream`?

```console
$ . /etc/os-release
$ echo $NAME
CentOS Stream
```

kenhys: Thanks,
It should use $NAME in this context.

kenhys:
I'll check CentOS Stream 8 case.

kenhys:
got it. fixed missing double quote in the script.

ashie (approved):

PR #326, opened 2021-08-16
rpm: add helper script to manage repository

kenhys:
TODO: it follows deb, too


kenhys:
I've added this script not to make an unexpected broken repository.

ashie:
> TODO: it follows deb, too

It's already resolved, isn't it?

PR #328, opened 2021-08-18
deb: support Debian 11 (bullseye)

kenhys:
Debian 11 "bullseye" released

https://www.debian.org/News/2021/20210814

### DIFF
--- a/td-agent-apt-source/apt/debian-bullseye/Dockerfile
+++ b/td-agent-apt-source/apt/debian-bullseye/Dockerfile
@@ -1,0 +1,17 @@
+FROM debian:bullseye
+
+RUN \
+  echo "debconf debconf/frontend select Noninteractive" | \
+    debconf-set-selections
+
+ARG DEBUG
+
+RUN \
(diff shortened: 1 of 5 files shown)


cosmo0920:
I confirmed that debian-bullseye amd64 package can be built with this change but built patch is changed in `td-agent/apt/repositories/debian/pool/n/a/main/t/td-agent`.
Note that debian-buster package will be written in `td-agent/apt/repositories/debian/pool/buster/main/t/td-agent`.
Is this change expected?

cosmo0920:
Hmm.., my box docker environment is still pointed at old bullseye image. I re-pulled debian:bullseye image...and testing for this PR.

cosmo0920 (changes requested):
We should add td-agent-apt-source as well:
https://github.com/fluent-plugins-nursery/fluent-package-builder/tree/master/td-agent-apt-source/apt

kenhys:
I've added apt-source too.

cosmo0920 (approved):
LGTM

PR #332, opened 2021-08-30
Bundle murmurhash for kafka plugin users

kenhys:
With it, you can keep compatibility with Java-based Kafka client behavior.
It enables you to choose murmur32 for fluent-plugin-kafka.


cosmo0920:
Sorry for my typo.

> It enables you to choose murmur32 for fluent-plugin-kafka.

murmur32 is not correct, murmur2 is correct hash algorithm name.

ashie (approved):

Issue #342, opened 2021-10-27
Log rotation on Windows


ashie:
From https://github.com/fluent/fluentd/issues/3538

### Describe the bug

The log file (C:\opt\td-agent\td-agent.log) is increasing continuously, how to put a limit ??

There is some configuration like file rotate and there is a command however we have a fluentd running as windows service, so if there is any configuration could you please suggest either in conf file or while running the fluentd service from powershell.

### To Reproduce

Keep fluentd service running and process different messages the file size keeps increasing.

### Expected behavior

The old contents in the file should be deleted after a certain limit is reached.

### Your Environment

```markdown
- Fluentd version:
- TD Agent version:
- Operating system:
- Kernel version:
```

### Your Configuration

```apache
Normal configuration
```

### Your Error Log

```shell
No Errors it's physical size issue.
```

kenhys:
@harsh288 

No, It's not working.

As I noted [before](https://github.com/fluent/fluentd/issues/3538#issuecomment-953425857), fluentd 1.12.3 will not accept rotate settings.

You can ensure by launch fluentd manually: (instead of launching as a service)

rotate.conf
```
<system>
  <log>
    rotate_age 10
    </log>
</system>
```

You can see a warning message.

```
> fluentd -c rotate.conf
2021-11-01 16:41:58 +0900 [info]: starting fluentd-1.12.3 pid=14780 ruby="2.7.3"
2021-11-01 16:41:58 +0900 [info]: spawn command to main:  cmdline=["C:/Ruby27-x64/bin/ruby.exe", "-Eutf-8", "C:/Ruby27-x64/bin/fluentd", "-c", ".\\rotate.conf", "--under-supervisor"]
2021-11-01 16:42:00 +0900 [warn]: parameter 'rotate_age' in <log>
  rotate_age 10
</log> is not used.
```

One thing you need to do is: upgrade Fluentd.

daipom:
@tivanov-qb  I see `rotate_age 0` setting on the image. Is it intended?

tivanov-qb:
@daipom i was trying to get the rotation working and appearently this is what I had in the file when I took the screenshot. You can also ignore the comment for 50MB, as you can see it is set to Kbs, but does not work :(

daipom:
@tivanov-qb Could you try removing `rotate_age 0` setting?

PR #343, opened 2021-11-02
Enable minimum log rotation configuration for windows

kenhys:
Closes: #342


kenhys:
After executing `yum groupinstall -y ${quiet} "Development Tools"`, yum install tend to fail.
On local environment, it doesn't happen.

ashie:
> After executing `yum groupinstall -y ${quiet} "Development Tools"`, yum install tend to fail. On local environment, it doesn't happen.

Hmm, it doesn't reproduce also on my local environment.

PR #344, opened 2021-11-02
deb: remove needless update conf hook

kenhys:
Since td-agent 4.0.0, td-agent.conf.tmpl was moved from
/etc/td-agent/td-agent.conf.tmpl to
/opt/td-agent/share/td-agent.conf.tmpl.

It means that td-agent.conf.tmpl was not registered in
Conffiles: entries anymore. (It's not under /etc)
Then, old_md5sum becomes always empty, backward compatibility
is triggered.

```
  Unpacking td-agent (4.2.0-1) over (4.1.1-1) ...
  Setting up td-agent (4.2.0-1) ...
  Updating conffile /etc/td-agent/td-agent.conf ...
  Processing triggers for man-db (2.9.4-2) ...
```

As a result, td-agent.conf file will be overridden. (It's no
harm because the content is same.)

In this commit, check template file checksum correctly.

Signed-off-by: Kentaro Hayashi <hayashi@clear-code.com>

### DIFF
--- a/td-agent/templates/package-scripts/td-agent/deb/postinst
+++ b/td-agent/templates/package-scripts/td-agent/deb/postinst
@@ -28,48 +28,11 @@ fixperms() {
         dpkg-statoverride --update --add <%= project_name %> <%= project_name %> 0755 /var/log/<%= project_name %>
 }
 
-update_conffile() {
-    CONFFILE="$1"
-    TMPL="$2"
-
-    if [ -e "$CONFFILE" ]; then
-        md5sum="`md5sum \"$CONFFILE\" | sed -e \"s/ .*//\"`"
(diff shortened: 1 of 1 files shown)


kenhys:
CentOS 7 aarch64 ci fials.

It may be a bit strange. is this package size valid for?

```
================================================================================
 Package   Arch    Version        Repository                               Size
================================================================================
Installing:
 td-agent  aarch64 4.2.0-1.el7    /td-agent-4.2.0-1.el7.aarch64            68 M
 td-agent-debuginfo
           aarch64 4.2.0-1.el7    /td-agent-debuginfo-4.2.0-1.el7.aarch64  76 M
```

ashie:
Is copying the template to /etc/td-agent/td-agent.conf still needed?
AFAIK, we already removed the equivalent code for RPM.

kenhys:
Oops, 

> Is copying the template to /etc/td-agent/td-agent.conf still needed?

It should be removed.

ashie (approved):

PR #349, opened 2021-11-29
GitHub Actions: Don't run on push events for branches

ashie:


ashie:
Without this, CI is always run not only for pull_request event but also for push event, it's too wasted.

kenhys (approved):
LGTM,

PR #350, opened 2021-11-30
Fix build error on Debian Bullseye

fujimotos:
qemu-user-static on Ubuntu 20.04 contains a bug that crashes arm64
Debian Bullseye guests:

     https://bugs.launchpad.net/qemu/+bug/1749393

This patch works around that issue, by installing the required qemu
binaries (in particular `qemu-aarch64-static`) using
"multiarch/qemu-user-static" image.

Signed-off-by: Fujimoto Seiji <fujimoto@ceptord.net>


ashie:
Thanks!

PR #357, opened 2022-01-14
yum: use rockylinux as base build image

kenhys:
It stores RPMs under td-agent/yum/repositories/centos for
compatibility.

TODO: support armv64v8/rockylinux:8 as well. It is not accessible from my
environment :-<, It will be handled in another PR.

### DIFF
--- a/td-agent/yum/rockylinux-8-aarch64/from
+++ b/td-agent/yum/rockylinux-8-aarch64/from
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


lib/yum/build.sh
```diff
@@ -40,6 +40,10 @@ if [ -f /etc/os-release ]; then
       # The ID is translated for repository layout
       distribution=amazon
       ;;
+    rocky)
+      # The ID is translated as transitional layout
+      distribution=centos
```
ashie, line 45: I think it's not needed to keep compatibility.
I prefer using the name as is (rocky) to avoid confusion.

ashie (changes requested):

lib/yum/build.sh
kenhys: Ok, just removed.

cosmo0920 (changes requested):
The correct name is rokcylinux not rock**e**ylinux.

kenhys:
Oops, thanks.
Fixed now.

ashie (approved):

cosmo0920:
🎉

PR #358, opened 2022-01-17
Bundle calyptia related gems

### LINKED ISSUE #333: Bundle Calyptia service related gems and plugins
Calyptia is planned to publish service for monitoring.

Currently, our created gems are as follows:

* cmetrics-ruby (cmetrics as actual gem name): https://github.com/calyptia/cmetrics-ruby
* fluent-plugin-metrics-cmetrics: https://github.com/calyptia/fluent-plugin-metrics-cmetrics
* fluent-plugin-calyptia-monitoring: https://github.com/calyptia/fluent-plugin-calyptia-monitoring

They are used for our new Calyptia Monitoring service.
Two of Fluentd plugins are ordinary Fluentd plugins.
But, cmetrics-ruby requests to use cmake 3.13 or later for TARGETS in subdirectories feature.
It had been supported in cmake 3.13 or later.
ref: https://cmake.org/cmake/help/v3.13/command/install.html#targets

This could be headache for old platforms including Ubuntu 16.04, Ubuntu 18.04, CentOS 7, and CentOS 6(Not tested yet).
For Ubuntu, using kitware apt repository can solve this headache and, for CentOS 7, using EPEL repository can solve it.

But, for CentOS 6, there is no cmake3 3.13 or later package is provided even if in EPEL.
So, if we bundle these Calyptia related gems on CentOS 6 platform, we might ought to create 3rd party package for cmake 3.13 or later....
CMake 3 compilation requests to use C++11 ready gcc compiler toolchain.
So, we also ought to build 3rd party package for CentOS 6 to build cmake 3 package if we proceed to bundle calyptia related gems. Or, Only for CentOS 6, nothing bundling for Calyptia related gems is also reasonable for us because CentOS 6 is already EOL.

FYI: `fluent-plugin-calyptia-monitoring` has a configuration generator command which is `calyptia-config-generator`: https://github.com/calyptia/fluent-plugin-calyptia-monitoring#calyptia-monitoring-api-config-generator

cosmo0920:
Closes #333

nokogiri v1.13 start to depends on mini_portile v2.7.1 and provides fat gems for Linux AArch64 platform.
Now, it's time to bundle Calyptia related gems and plugins into fluent-package-builder building system.


cosmo0920:
How do we treat Windows package building failure?

cosmo0920:
I got succeeded to create Windows msi installer by `rake msi:selfbuild` instead of `rake msi:build`.

ashie:
> How do we treat Windows package building failure?

It's not related with this.
Should be fixed in another PR.

ashie (approved):

cosmo0920:
@kenhys @ashie Can we merge this PR?

cosmo0920:
It would be nice if we can merge this PR before #360 is merged.

cosmo0920:
Thanks!

PR #365, opened 2022-02-02
Add fluent-plugin-windows-exporter

### LINKED ISSUE #355: Add a new windows-exporter plugin to td-agent
### Proposal

* Bundle fluent-plugin-wndows-exporter to td-agent v4.4.0
  https://github.com/fluent-plugins-nursery/fluent-plugin-windows-exporter

### Rationale

* This plugin allows users to collect system metrics on Windows using Fluentd. 
* It is a very "comprehensive" plugin. Currently, it supports 90+ items of metrics.
* With this bundled, Windows users now can use Fluentd as the single source for observability, saving the trouble of installing multiple daemons (Prometheus and collectd etc.)

### Technical consideration
 
* This plugin is fully written in Ruby (we use fiddle to interact with Win32 API, so no C lib is required).
* The only dependency is [bindata](https://github.com/dmendel/bindata), which is also just a Ruby module.
* So there must be very little trouble to include this plugin.

fujimotos:
Fixes #355.

Windows Exporter plugin is a system metrics plugin for Win32 platforms.
It can capture a wide range of metrics (CPU, memory, and network etc.),
and is very convenient for monitoring Windows nodes.

* https://github.com/fluent-plugins-nursery/fluent-plugin-windows-exporter
* https://rubygems.org/gems/fluent-plugin-windows-exporter

The plugin enables Fluentd to act as a replacement of general system
monitoring solution (such as collectd or Prometheus), and should be
very useful for Windows administrators.

Signed-off-by: Fujimoto Seiji <fujimoto@ceptord.net>

### DIFF
--- a/td-agent/Gemfile
+++ b/td-agent/Gemfile
@@ -84,6 +84,7 @@ gem "winevt_c", "0.9.3", platforms: windows_platforms
 gem "win32-eventlog", "0.6.7", platforms: windows_platforms
 gem "fluent-plugin-parser-winevt_xml", "0.2.4", platforms: windows_platforms
 gem "fluent-plugin-windows-eventlog", "0.8.1", platforms: windows_platforms
+gem "fluent-plugin-windows-exporter", "1.0.0", platforms: windows_platforms
 
 not_windows_platforms = [:ruby]
 gem "rdkafka", "0.11.1", platforms: not_windows_platforms
--- a/td-agent/Gemfile.lock
+++ b/td-agent/Gemfile.lock
@@ -69,6 +69,7 @@ GEM
     console (1.14.0)
       fiber-local
     cool.io (1.7.1)
+    cool.io (1.7.1-x64-mingw32)
     digest-crc (0.6.4)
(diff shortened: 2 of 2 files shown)


td-agent/Gemfile
```diff
@@ -70,6 +70,7 @@ gem "fluent-plugin-metrics-cmetrics", "0.1.1"
 gem "fluent-plugin-calyptia-monitoring", "0.1.3"
 gem "opensearch-ruby", "1.0.0"
 gem "fluent-plugin-opensearch", "1.0.0"
+gem "fluent-plugin-windows-exporter", "1.0.0"
```
ashie, line 73: Tests failed on non-windows platforms.
Please add `platforms: windows_platforms` (see the below block).

ashie (changes requested):

td-agent/Gemfile
fujimotos: Sorry! fixed by d26444498585035f9bb80261da1802629241b1d8.

Issue #370, opened 2022-02-11
td-agent RPM includes build-ids


Smasherr:
Due to https://github.com/chef/omnibus/pull/1050, td-agent has conflicts with gitlab-ce and an installation of td-agent results in

```
Error: Transaction test error:
  file /usr/lib/.build-id/3a/a6377e3b490d35f6b2e13af2a0a487992a4d7b from install of td-agent-4.3.0-1.el8.x86_64 conflicts with file from package gitlab-ce-14.7.1-ce.0.el8.x86_64
  file /usr/lib/.build-id/6b/074ebdad7c55f515e55faf0c859d96ce823c74 from install of td-agent-4.3.0-1.el8.x86_64 conflicts with file from package gitlab-ce-14.7.1-ce.0.el8.x86_64
  file /usr/lib/.build-id/7b/152f6fd3336942261a5c46bb46cf1355c8368f from install of td-agent-4.3.0-1.el8.x86_64 conflicts with file from package gitlab-ce-14.7.1-ce.0.el8.x86_64
  file /usr/lib/.build-id/86/47df0ea0bafe7bd29ac2c12584acdb9addb147 from install of td-agent-4.3.0-1.el8.x86_64 conflicts with file from package gitlab-ce-14.7.1-ce.0.el8.x86_64
```

This project should apparently replace omnibus, but the [latest version of td-agent available at packages.treasuredata.com](https://s3.amazonaws.com/packages.treasuredata.com/4/redhat/8/x86_64/td-agent-4.3.0-1.el8.x86_64.rpm) still includes build-ids:

```
❯ tar tf td-agent-4.3.0-1.el8.x86_64.rpm | grep a6377e3b490d35f6b2e13af2a0a487992a4d7b
./usr/lib/.build-id/3a/a6377e3b490d35f6b2e13af2a0a487992a4d7b
``` 
Are there any plans to deploy a new version of td-agent build with this new packager to packages.treasuredata.com? Or is there another repository that should be used?

PR #371, opened 2022-02-14
yum: generate build_id in -debuginfo

kenhys:
build_id links may be useful for debug but there is a case that
build_id conflicts with other packages. (e.g. gitlab-ce)

Thus "_build_id_links compat" may not be suitable, so change it to
traditional alldebug.

Before: 
  td-agent contains /usr/lib/debug/.build-id/xx/yy
After: 
  td-agent-debuginfo contains /usr/lib/debug/.build-id/xx/yy

NOTE: rpm --excludepath may be a workaround to install, but it is not
so good.

Closes: #370


ashie (approved):

Smasherr:
Thanks for resolving this!

PR #381, opened 2022-04-25
Support Ubuntu 22.04

kenhys:
This draft PR is derived from https://github.com/fluent/fluentd/discussions/3704 and push a commit on https://github.com/fluent/fluent-package-builder/pull/378
to attempt to identify Ubuntu 22.04 buildability. (not ready for production)

* Use newer ruby only for Ubuntu 22.04 because of supporting OpenSSL 3.x

KNOWN ISSUE: dh_shlibdeps causes an error.

```
dpkg-shlibdeps: error: no dependency information found for /build/td-agent-4.3.1/debian/tmp/opt/td-agent/lib/libruby.so.3.1 (used by debian/
```

PR #384, opened 2022-05-02
Revise apt-source package

kenhys:
This PR is a follow-up of #381.
It is aimed to eliminate deprecated apt-key(8) usage.


ashie:
@kenhys Could you resolve the conflict?

ashie (approved):

PR #387, opened 2022-05-09
msi: Replace OpenSSL with MinGW's latest one to fix memory leak

### LINKED ISSUE #374: Memory leak when using TLS on Windows
Recently we received a bug report from a customer that Windows version of td-agent causes memory leak.
It's reproducible with td-agent 4.3.0.
After I investigated it, I found out that just using TLS causes this issue.
In addition, Ruby doesn't recognize it. I confirmed it by [fluent-plugin-sigdump](https://github.com/fluent-plugins-nursery/fluent-plugin-sigdump).

fluent-plugin-windows-eventlog and certstore_c also don't have major memory leak, I confirmed it by Dr.Memory and reviewing the code.
While testing cerstore_c by Dr.Memory, OpenSSL related leaks are detected. So that the origin of this memory leaks might be Ruby (openssl gem) or OpenSSL. We don't yet understand the detail due to lack of debug symbols.

Although the memory leak by one communication is tiny, it's not ignorable. Because out_forward plugin of Fluentd discards a socket every time after finishing a communication and recreate it by default, so that memory leaks are accumulated.

It can be mitigate by `keepalive` option of out_forward, but we need to fix the root cause to allow long-life operation.

This issue doesn't reproduce when I use Ruby 2.4.

ashie:
Fix #374

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -401,7 +401,7 @@ class BuildTask
     namespace :build do
       desc "Install jemalloc"
       task :jemalloc => [:"download:jemalloc"] do
-        build_jemalloc unless windows?
+        build_jemalloc
       end
 
       desc "Install OpenSSL"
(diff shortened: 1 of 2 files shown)


kenhys (approved):
Could you amend a commit message to include a summary of #374 for the record?

ashie:
Done: 0a4aa1b0111418efbbcdf0409e2fa1abc7d36871

PR #389, opened 2022-05-23
Update nokogiri to 1.13.6, and install it only on Windows

ashie:
Only fluent-plugin-windows-eventlog and fluent-plugin-parser-winevt_xml
are explicitly depdends on nokogiri.
Because nokogiri is often reported vulnerabilities, it should be installed
and updated by td-agent-gem if users really require it.

TODO: Probably we need to install rexml or something other for aws-sdk-core
with Ruby 3.x:

https://github.com/aws/aws-sdk-ruby/blob/64e9ea97c318cb7a10fdf7c0979307a772f06b51/gems/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb#L70


kenhys (approved):
LGTM.

cosmo0920 (approved):
Basically, until we will use Ruby 3.x, this change should be correct.
For Ubuntu 22.04 package, we need to install nokogiri for aws-sdk-core?
Shouldn't we install for Ruby 3.x environment?

Should we leave it in a comment, or file an issue for Ruby 3.x upgrade?

ashie:
> For Ubuntu 22.04 package, we need to install nokogiri for aws-sdk-core? Should we leave it in a comment, or file an issue for Ruby 3.x upgrade?

Probably we don't need to do anything, rexml will be installed automatically.
Although rexml is no longer "default gem" since Ruby 3.0, it's still "bundled gem".
I confirmed that ubuntu-jammy package built by this branch includes rexml gem.

cosmo0920:
Noted. Thanks!

PR #403, opened 2022-06-28
manage-fluent-repositories.sh: Further automation of creating RPM repo

ashie:
In addition, fix wrong usage in help.

### DIFF
--- a/td-agent/manage-fluent-repositories.sh
+++ b/td-agent/manage-fluent-repositories.sh
@@ -17,11 +17,11 @@ Usage:
   $0 COMMAND FLUENT_RELEASE_PROFILE FLUENT_RELEASE_DIR FLUENT_PACKAGE_VERSION"
 
 Example:
-  $ $0 ls release-td-agent
-  $ $0 download release-td-agent /tmp/td-agent-release
-  $ $0 deb release-td-agent /tmp/td-agent-release 4.2.0
-  $ $0 rpm release-td-agent /tmp/td-agent-release 4.2.0
-  $ $0 upload release-td-agent /tmp/td-agent-release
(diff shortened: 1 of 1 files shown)


td-agent/manage-fluent-repositories.sh
```diff
@@ -122,13 +122,16 @@ EOF
...
-	# sign rpm repository
-	find $FLUENT_RELEASE_DIR/4 -name 'repomd.xml'
-	for f in `find $FLUENT_RELEASE_DIR/4 -name 'repomd.xml'`; do
-	    if [ -f "$f.asc" ]; then
-		rm -f $f.asc
+	# update & sign rpm repository
+	repodirs=`find "${FLUENT_RELEASE_DIR}" -regex "^${FLUENT_RELEASE_DIR}/4/\(redhat\|amazon\)/[2789]$"`
+	for repodir in $repodirs; do
+	    createrepo_c -v "${repodir}"
+
```
kenhys, line 129: In the previous versions, repodata/repomd.xml is generated under each x86_64 or aarch64.
Is it intentional change?, If so and it should be changed, it's ok for me.

kenhys: I mean: 

Before: td-agent-release/4/redhat/8/x86_64/repodata/repomd.xml
After:  td-agent-release/4/redhat/8/repodata/repomd.xml

ashie: It's not intended. I'll fix it.

ashie: Fixed in 63c1e084fe324c1c90a8c7686107710cddc8d672

PR #418, opened 2022-08-29
Enable upgrade test for Rocky Linux & AlmaLinux

ashie:


ashie:
The upgrade test is for upgrading from td-agent 3, but there is no td-agent 3 package for RHEL 8 & 9.
So I close this.

PR #420, opened 2022-10-25
v4.4.2

ashie:


ashie:
We postpone releasing it until the detail of a critical vulnerability of OpenSSL v3 is unveiled and we complete checking it:

https://mta.openssl.org/pipermail/openssl-announce/2022-October/000238.html

(Probably it doesn't affect to our product though)

ashie:
> We postpone releasing it until the detail of a critical vulnerability of OpenSSL v3 is unveiled and we complete checking it:

The detail is published:
https://mta.openssl.org/pipermail/openssl-announce/2022-November/000243.html

OpenSSL v1.1.1 isn't affected.
We are going to release td-agent 4.4.2 at this weekend.

fujimotos:
@ashie FYI, I'm planning to spend a couple of hours testing the build
before the release. 

I _really_ don't want any more regressions, so will test both on Linux and
Windows with several patterns of configurations. I'm going to post more
updates here.

ashie:
CI stalls on Windows, it seems that there are 2 causes.

1. Building the docker image is stalled due to using older MSYS2 as we experienced several times before.
2. I'm not sure why but Bundler 2.3.24 always stalls while installing gems on Windows. So I downgraded it to 2.3.18 which is used in the previous version, It successfully complete the installation.

fujimotos:
OK. I checked this build with my dev machine as well ass
RHEL 8.6. I think this is realeasable.

ashie:
I've uploaded v4.4.2 packages.
https://td-agent-package-browser.herokuapp.com/4

PR #426, opened 2023-01-27
Replace File.exists? with File.exist?

ashie:
File.exists? is removed as of Ruby 3.2


daipom (approved):
LGTM.
Thanks for the fix!

ashie:
Waiting CI...

PR #283, opened 2021-03-18
Remove needless dependencies on RHEL/CentOS 6, again

ashie:
5c905844edf9e4a2db556ef9ba16f4e3c4ea9c76 lacks filters for provides.

PR #430, opened 2023-02-09
Drop support for CentOS 6

kou:
https://github.com/fluent/fluent-package-builder/pull/283#issuecomment-801672273 said that we will maintain CentOS 6/RHEL 6 until 2024-06-30 but it's difficult for us because we can't prepare CI for CentOS 6/RHEL 6.


ashie:
CentOS 6 was remained mainly for a particular customer, but there is no contact about this matter from them for about two years. And ELS is remained only about one year. So I think it's time to drop it.

PR #436, opened 2023-03-31
Update fluentd to v1.16.0

ashie:


ashie:
Install test failed on Windows:
https://github.com/fluent/fluent-package-builder/actions/runs/4571636806/jobs/8070117936?pr=436
```
Succeeded to find td-agent.conf
Failed to find td-agent-0.log: <>
```

ashie:
Hmm, cannot reproduce on my local environment...

daipom:
On my local Windows, the supervisor initialization looks like failing, and the supervisor and worker processes are not launched.

This seems to be caused by the following fix.

* https://github.com/fluent/fluentd/pull/4087

With the default config, the following warning log about the secondary occurs.

```
[warn]: [output_td] Use different plugin for secondary. Check the plugin works with primary like secondary_file primary="Fluent::Plugin::TreasureDataLogOutput" secondary="Fluent::Plugin::FileOutput"
```

This process looks like the cause.

When I checked the error as follows,

```ruby
begin
  if (@secondary.class != SecondaryFileOutput) &&
     (self.class != @secondary.class) &&
     (@custom_format || @secondary.implement?(:custom_format))
    log.warn "Use different plugin for secondary. Check the plugin works with primary like secondary_file", primary: self.class.to_s, secondary: @secondary.class.to_s
  end
rescue => e
  $log.error "Error: ", e
end
```

I found this raises the following error.

```
uninitialized constant Fluent::Plugin::Output::SecondaryFileOutput
```

daipom:
I don't see why this phenomenon does not occur when running Fluentd normally...

I can confirm this phenomenon occurs on macOS as well.
Perhaps this phenomenon is not specific to Windows.

daipom:
> I don't see why this phenomenon does not occur when running Fluentd normally...

This phenomenon is not specific to this package.
This error does not occur if `out_secondary_file` is set once before setting a secondary other than it.

I have made a PR to fix this.

* https://github.com/fluent/fluentd/pull/4124

ashie:
> Hmm, cannot reproduce on my local environment...

It's my mistake. I built the package with v1.15.3.

ashie:
>     * [Fix NameError of `SecondaryFileOutput` when setting secondary other than `out_secondary_file` fluentd#4124](https://github.com/fluent/fluentd/pull/4124)

The issue has been resolved by this.
Now it seems enough for testing, so I merge this.
(We should replace with fluentd v1.16.1 until we release the next version though.)

PR #439, opened 2023-04-03
CI: Don't trigger build actions for Arm platform automatically

ashie:
They take too long time to complete on CI environment since they use QEMU. Due to this they often time out.

Instead trigger them manually by `workflow_dispatch`. 
In addition, I'm considering to prepare self-hosted runner for them.


ashie:
Thanks!

PR #442, opened 2023-04-13
deb: Remove needless "#{PACKAGE_NAME}" prefix

kou:
"debian/copyright" and "debian/source/include-binaries" are generated to "debian/copyright" not "#{PACKAGE_NAME}/debian/copyright" and so on. If we use wrong path to `file`, we always regenerate these files. It slows down `rake apt:build`.


ashie:
Could you follow DCO?: https://github.com/fluent/fluent-package-builder/pull/442/checks?check_run_id=12714000712

PR #447, opened 2023-04-19
td-agent v4.5.0

ashie:

PR #451, opened 2023-05-02
td-agent-v4: Replace File.exists? with File.exist?

ashie:
File.exists? is removed as of Ruby 3.2


ashie:
Backport from master to td-agent-v4 branch.
It's already tested for td-agent-v4 branch in #447.
To narrow the scope of #447, I've separated the commit to this PR.

PR #452, opened 2023-05-02
td-agent-v4: deb: Remove needless "#{PACKAGE_NAME}" prefix (#442)

ashie:
"debian/copyright" and "debian/source/include-binaries" are generated to "debian/copyright" not "#{PACKAGE_NAME}/debian/copyright" and so on. If we use wrong path to `file`, we always regenerate these files. It slows down `rake apt:build`.

### DIFF
--- a/td-agent/Rakefile
+++ b/td-agent/Rakefile
@@ -1192,12 +1192,12 @@ class LinuxPackageTask < PackageTask
       File.join("..", path)
     end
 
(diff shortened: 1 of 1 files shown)


ashie:
Backport from master to td-agent-v4 branch.
It's already tested for td-agent-v4 branch in #447.
To narrow the scope of #447, I've separated the commit to this PR.

PR #455, opened 2023-05-02
td-agent-v4: Drop Debian buster

ashie:
It reaches EOL already.


ashie:
It's already tested for td-agent-v4 branch in https://github.com/fluent/fluent-package-builder/pull/447.
To narrow the scope of https://github.com/fluent/fluent-package-builder/pull/447, I've separated the commit to this PR.

PR #458, opened 2023-05-08
CI: Fix target branch for td-agent v4

ashie:


ashie:
I tested it in my fork but still failed: https://github.com/ashie/fluent-package-builder/tree/td-agent-v4

daipom:
Thanks for this fix.

Don't we need to specify the branch at `actions/checkout`?

```yml
- uses: actions/checkout@master
  with:
    ref: td-agent-v4
```

ashie:
> I tested it in my fork but still failed: https://github.com/ashie/fluent-package-builder/tree/td-agent-v4

This is because I didn't push tags to my fork.
It's needed for package size test: https://github.com/fluent/fluent-package-builder/blob/33f482c8fdcc9ebb17ed6b03aa4211707ac6a64a/fluent-package/yum/pkgsize-test.sh#L21-L23

ashie:
> Don't we need to specify the branch at `actions/checkout`?
> 
> ```yaml
> - uses: actions/checkout@master
>   with:
>     ref: td-agent-v4
> ```

It's performed against committed revision, so it's not needed in usual.

daipom:
Oh, I see! Thanks!

PR #469, opened 2023-05-18
msi: stop customizing icons for file browser

daipom:
We should use the default wix icons for these.


daipom:
2023-05-18_15h58_24

before

2023-05-18_16h20_47

after

2023-05-18_15h58_09

ashie:
😱

ashie (approved):

ashie:
Thanks!

daipom:
Thanks!

PR #470, opened 2023-05-19
msi: update resources

daipom:


daipom:
I have updated these images by using logos from https://docs.fluentd.org/quickstart/logo

daipom:
I have used GIMP to convert `.png` to `.ico`.

daipom:
I have made the size of the `.ico` file `32 * 32`.
I can't find any rules about `ARPPRODUCTICON` of WiX.
According to [Microsoft document](https://learn.microsoft.com/en-us/windows/win32/uxguide/vis-icons#size-requirements), it seems that we should have `32*32` size.

daipom:
Here are the results I have tried with my fork.

2023-05-18_23h26_12
2023-05-18_23h26_23
2023-05-18_23h27_07
2023-05-18_23h29_43
2023-05-18_23h31_58

daipom:
Thanks for your review!

PR #471, opened 2023-05-19
msi: update ManufacturerName

daipom:


daipom:
2023-05-18_23h29_43

(This image includes the fixes of #470)

PR #474, opened 2023-05-19
dmg: simplify distribution.xml

daipom:


daipom:
I can't see the intent of the current implementation.
If there is no intent, it would be better to keep it as simple as this.

kenhys:
It may be obvious for macOS developer, but it is good practice to mention  "why it is reasonable change".
Indicating Distribution XML Reference is one example to achive it:
 https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW9

daipom:
Sorry for my lack of explanation.

> it is good practice to mention "why it is reasonable change".

Understood!
Thanks for sharing the document.

I'm not sure if this change is reasonable. 
I read this document, but I didn't understand the current code meaning. 

If anyone knows the intent of the current code, I would like to know.

It is OK that we don't know, so let's not change it for now.

kenhys:
I'm not against the effort to simplify it.

> I read this document, but I didn't understand the current code meaning.

I have a concern that it may tend to repeat above situation if we didn't explain it with proper context .
(I mean that  when you can explain how we can simplify it to some extent, e.g. According to ref xxx, it should be .... and so on, it is ready to merge it)

daipom:
> I'm not against the effort to simplify it.
> 
> > I read this document, but I didn't understand the current code meaning.
> 
> I have a concern that it may tend to repeat above situation if we didn't explain it with proper context . (I mean that when you can explain how we can simplify it to some extent, e.g. According to ref xxx, it should be .... and so on, it is ready to merge it)

Understood! I agree!

I will add information that I know about this later!

Now I noticed we set `customize="never"`, so we can't see `choice-outline` in the installer.

daipom:
> > I have a concern that it may tend to repeat above situation if we didn't explain it with proper context .
>
> I will add information that I know about this later!

Why I thought this fix was necessary:

According to https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW20, I know the following points.

* `choices-outline` contains one or more `line`, and `line` contains zero or more `line`. These specifies the installation-choice hierarchy.
* `line` specifies an id of `choice`, which contains zero or more `pkg-ref`. `pkg-ref` specifies an identifier of a component package.

td-agent has only one component, so we don't need to customize these choices. (We set `customize="never"`)
From the above points, it would be natural to set only one `line` and one `choice` like this fix.

However,  I'm concerned that I am not sure what the current code intends.

ashie (approved):

daipom:
@kenhys @ashie Thanks for your review!

PR #479, opened 2023-05-24
dmg: change identifier

daipom:
It should be fine as long as it is unique.
`org.fluent.fluentd` would be appropriate.

### DIFF
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -1485,7 +1485,7 @@ class MacOSPackageTask
       sh("pkgbuild",
          "--root", STAGING_DIR,
          "--component-plist", File.join("resources", "pkg", "#{SERVICE_NAME}.plist"),
-         "--identifier", "com.treasuredata.tdagent",
+         "--identifier", "org.fluent.fluentd",
          "--version", @version,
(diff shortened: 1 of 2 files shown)


daipom:
`$ man pkgbuild`

>     --identifier pkg-identifier
>             Specify a unique identifier for this package. The OS X Installer recognizes a package as being an upgrade to an already-installed package only
>             if the package identifiers match, so it is advisable to set a meaningful, consistent identifier when you build the package.  pkgbuild will
>             infer an identifier when building a package from a single component, but will fail otherwise if the identifier has not been set.

https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW11

`id` attribute of `pkg-ref`

> The value of this attribute is usually the same as the component package’s identifier,

daipom:
Waiting CI...

daipom:
All success.

ashie (approved):

ashie:
Thanks!

daipom:
Thanks!

PR #480, opened 2023-05-24
dmg: fix bin scripts paths

daipom:
Change both `/opt/fluent/usr/sbin/` and `/opt/fluent/usr/bin/` paths
to `/opt/fluent/sbin/` for simplicity.

Note: we can't put them into `/opt/fluent/bin/` because the filenames
conflict with internal Fluentd gems.

### DIFF
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -517,7 +517,9 @@ class BuildTask
         scripts.each do |script|
           src = template_path("#{script}.erb")
           dest = if macos?
-                   File.join(STAGING_DIR, File.join("opt", PACKAGE_DIR, script))
+                   # We should avoid touching `/usr/sbin/` because we
+                   # don't prepare the uninstall feature for macOS.
+                   File.join(td_agent_staging_dir, "sbin", File.basename(script))
                  else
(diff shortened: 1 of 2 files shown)


daipom:
Waiting CI...

daipom:
Oops, I noticed that `/opt/fluent/usr/sbin/fluent-gem` would overwrite `/opt/fluent/bin/fluent-gem`.

Separately with this PR, we have the idea to overwrite it since it is confusing.

After that fix is made, we may as well stay with this.

daipom:
> Oops, I noticed that `/opt/fluent/usr/sbin/fluent-gem` would overwrite `/opt/fluent/bin/fluent-gem`.
> 
> Separately with this PR, we have the idea to overwrite it since it is confusing.
> 
> After that fix is made, we may as well stay with this.

Overwriting `/opt/fluent/bin/` is difficult because the gem manages those files. (For example, updating the gem will update those files.)

So I have changed the path to `/opt/fluent/sbin/`.

ashie (approved):

daipom:
Thanks for your review!

PR #482, opened 2023-05-24
dmg: fix service name

daipom:


daipom:
The tests of macOS succeed.

PR #484, opened 2023-05-24
msi: migrate td-agent*.bat

kenhys:
msi: migrate td-agent*.bat

Before:

c:\opt\td-agent\bin\td-agent.bat
c:\opt\td-agent\bin\td-agent-gem.bat

After:

c:\opt\td-agent\fluentd.bat
c:\opt\td-agent\fluent-gem.bat

If td-agent-gem.bat was renamed to fluent-gem.bat, it conflict with
c:\opt\td-agent\fluent-gem.bat which fluentd gem provides.
Not to cause such a situation, put them under c:\opt\td-agent.

To keep compatibility, td-agent.bat and td-agent-gem are provided
as hardlink.

### DIFF
--- a/fluent-package/msi/assets/fluent-package-prompt.bat
+++ b/fluent-package/msi/assets/fluent-package-prompt.bat
@@ -1,3 +1,4 @@
 @echo off
 set PATH="%~dp0\bin";%PATH%
+set PATH="%~dp0";%PATH%
 title Fluent Package Command Prompt
--- a/fluent-package/msi/source.wxs
+++ b/fluent-package/msi/source.wxs
@@ -93,19 +93,15 @@
           </Component>
         </Directory>
       </Directory>
-      <Directory Id="TDAgentBinDir"
-                 Name="bin"
-                 ComponentGuidGenerationSeed="3ad15499-f8a4-471b-bc7f-16aa4abfd894">
-        <Component Id="TDAgentBat">
-          <File Name="td-agent.bat"
-                KeyPath="yes"
-                Source="$(var.ProjectSourceDir)\bin\td-agent.bat" />
-          <ServiceControl Id="ServiceControlFluentdWinSvc"
-                          Name="fluentdwinsvc"
-                          Stop="uninstall"
-                          Remove="uninstall" />
-        </Component>
-      </Directory>
+      <Component Id="FluentdBat" Guid="261158e8-7b22-48ff-b7eb-9f8296f30236">
+        <File Name="fluentd.bat"
+              KeyPath="yes"
+              Source="$(var.ProjectSourceDir)\fluentd.bat" />
+        <ServiceControl Id="ServiceControlFluentdWinSvc"
+                        Name="fluentdwinsvc"
+                        Stop="uninstall"
+                        Remove="uninstall" />
+      </Component>
     </DirectoryRef>
 
     <SetDirectory Id="WINDOWSVOLUME" Value="[WindowsVolume]" />
@@ -131,8 +127,8 @@
       <ComponentGroupRef Id="ProjectDir" />
       <ComponentRef Id="ApplicationShortcut" />
       <ComponentRef Id="TDAgentConf" />
-      <ComponentRef Id="TDAgentBat" />
       <ComponentRef Id="TDAgentAcl" />
+      <ComponentRef Id="FluentdBat" />
     </Feature>
 
     <!-- UI Stuff -->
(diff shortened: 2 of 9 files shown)


kenhys:
failure:: assets/fluentd.bat is overriden.

daipom (approved):
These diffs look good to me.
About the link files, I will confirm them after the installer is built.

ashie:
> So in this PR, changed to provide unique batch file - fluent.bat.

`fluent` means to wide nowaday. Although it's origianl name of fluentd, it means our organization name.
So that it might be ambiguous for users what the `fluent.bat` does.

How about placing `fluentd.bat` and `fluent-gem.bat` just under `/opt/td-agent(fluent)`, and add there to `$PATH`.
I think it has some merits.

* Placing execution files just under the application folder is quite usual on Windows.
* Although they have same name with the one under `/opt/td-agent/bin`, users can  easily recognize that the files just under `/opt/td-agent/` have higher priority.
* They won't overwritten by gem.

Or leaving td-agent.bat and td-agent-gem.bat as is and put symlinks with `fluent` prefix just under `/opt/td-aget(fluent)` might be better.

daipom:
@ashie  I see. I agree.

@kenhys About the link files:

```console
$ td-agent.lnk --version
td-agent 5.0.0 fluentd 1.16.1 (0a6d706a9cee5882d751b2cc6169696709df0134)

$ td-agent --version
'td-agent' は、内部コマンドまたは外部コマンド、
操作可能なプログラムまたはバッチ ファイルとして認識されていません。
```

daipom:
It seems that we can omit `.bat` but we can't omit `.lnk`.
I don't know this Windows specification well though,,,

ashie:
> NOTE: There is no need to provide td-agent-gem.bat as asset
> because it is almost same as fluent-gem.bat which is installed by
> fluentd gem.

It might be better to modify to set `GEM_HOME` and `GEM_PATH` like the one on GNU/Linux.
Setting them can suppress referring other directories unexpectedly configured by users.

kenhys:
> $ td-agent --version
'td-agent' は、内部コマンドまたは外部コマンド、
操作可能なプログラムまたはバッチ ファイルとして認識されていません。

Thank you for verifying it.

It seems that hardlink is better solution...

kenhys:
> How about placing fluentd.bat and fluent-gem.bat just under /opt/td-agent(fluent), and add there to $PATH.
I think it has some merits.

Putting these files under /opt/td-agent is a good approach. I'll follow it.

kenhys:
Even though without custom action, rollback occurs. checking...

kenhys:
> Even though without custom action, rollback occurs. checking...

[withoutca.log](https://github.com/fluent/fluent-package-builder/files/11561514/withoutca.log)

Terminated after property change:

> MSI (c) (88:1C) [14:50:00:813]: PROPERTY CHANGE: Adding dir8D9828541E9745DA0C06D1FC23D6C150 property. Its value is 'c:\opt\

fluent-package/msi/source.wxs
```diff
@@ -170,18 +172,71 @@
...
                   BinaryKey="WixCA"
                   DllEntry="WixQuietExec64"
                   Execute="deferred"
                   Return="check"
                   Impersonate="no" />
+
+    <Property Id="CreateCompatTdAgentBat" Value=" "/>
+    <CustomAction Id="SetCreateCompatTdAgentBat"
+                  Property="CreateCompatTdAgentBat"
+                  Value="&quot;cmd&quot; /c &quot;fsutil hardlink create[PROJECTLOCATION]bin\td-agent.bat [PROJECTLOCATION]fluentd.bat&quot;"/>
```
daipom, line 186:
```suggestion
                  Value="&quot;cmd&quot; /c &quot;fsutil hardlink create [PROJECTLOCATION]bin\td-agent.bat [PROJECTLOCATION]fluentd.bat&quot;"/>
```

kenhys:
Old bat is provided as hardlink.

image

kenhys:
checking with/without ComponentGuidGenerationSeed.

ashie (changes requested):
Could you modify also td-agent-prompt.bat to add top directory to `PATH`? The top folder should have higher priority than `bin`.

https://github.com/fluent/fluent-package-builder/blob/0d64c2819f22a129ab9023871b34cb9b911048d7/fluent-package/msi/assets/fluent-package-prompt.bat#L2

kenhys:
dead hardlink issue was fixed.

kenhys:
TODO: fix mismatched path issue

kenhys:
> `TODO:` fix mismatched path issue

Fixed it.

fluent-package/msi/assets/fluent-package-prompt.bat
```diff
@@ -1,3 +1,4 @@
 @echo off
 set PATH="%~dp0\bin";%PATH%
+set PATH="%~dp0";%PATH%
```
ashie, line 3: `%~dp0\bin` is still needed too.

ashie: For `fluent-cat`, `fluent-ctl` and so on.

kenhys: > %~dp0\bin is still needed too.

`%~dp0\bin` is not removed in this PR. Am i missing?

ashie: Sorry, I overlooked it.

daipom (approved):
I have confirmed this msi package works on my environment.

On fluent-package-prompt:

* Both `$ fluentd --version` and `$ td-agent --version` work expectedly.
* Both `$ fluent-gem environment` and `$ td-agent-gem environment` work expectedly.
* `$ fluent-ctl restart` works correctly.

On PowerShell:

* `$ Restart-Service fluentdwinsvc` works correctly.

Etc:

* It can be uninstalled correctly.

ashie:
I'm still checking...

ashie (approved):

ashie:
Thanks!

PR #485, opened 2023-05-25
CI: update package name to fluent-package

daipom:
This fixes the not found path to upload in apt.

    Warning: No files were found with the provided path:
      td-agent/apt/repositories. No artifacts will be uploaded.
    Warning: No files were found with the provided path:
      td-agent-apt-source/apt/repositories. No artifacts will be uploaded.


ashie:
Thanks!

PR #486, opened 2023-05-25
Use out_secondary_file for secondary output

daipom:
To solve the warning message:

    Use different plugin for secondary ...

### DIFF
--- a/fluent-package/apt/systemd-test/install-newly.sh
+++ b/fluent-package/apt/systemd-test/install-newly.sh
@@ -9,6 +9,9 @@ sudo apt install -V -y \
 
 systemctl status --no-pager fluentd
 
+sleep 3
+! grep -q -e '\[warn\]' -e '\[error\]' -e '\[fatal\]' /var/log/fluent/fluentd.log
+
 sudo apt remove -y fluent-package
 
 test -h /etc/systemd/system/fluentd.service
--- a/fluent-package/dmg/install-test.sh
+++ b/fluent-package/dmg/install-test.sh
@@ -13,6 +13,7 @@ hdiutil detach /Volumes/fluent-package
 
 sleep 10 # Wait for Fluentd to start up and run correctly.
 test $(sudo launchctl list fluentd | grep LastExitStatus | grep -oE "[0-9]+") = 0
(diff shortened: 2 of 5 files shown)


daipom:
This will change the output path as follows:

* before: `/var/log/td-agent/failed_records_0.log`
* after: `/var/log/td-agent/failed_records/dump.bin.0`

If this change is unacceptable, then we should not make this change.

daipom:
> This will change the output path as follows:
> 
>     * before: `/var/log/td-agent/failed_records_0.log`
> 
>     * after: `/var/log/td-agent/failed_records/dump.bin.0`
> 
> 
> If this change is unacceptable, then we should not make this change.

Note: `out_secondary_file` can't output the files with the same format as `out_file`.
Even if we use `base_name` as follows,  the path will be like `/var/log/td-agent/failed_records.0`.

```xml
    <secondary>
      @type secondary_file
      directory /var/log/td-agent
      basename failed_records
    </secondary>
```

ashie (approved):
Since the next release is major update, so it's good timing to change.

ashie (changes requested):
Sorry, I forgot to point out a thing.
Could you modify also fluent-package/templates/opt/fluent/share/td-agent.conf.tmpl?

daipom:
Oh, sorry. We need to fix it too. Thanks for your review.
I will fix it and confirm the behavior on each platform.

daipom:
* Waiting #525

daipom:
* Rebased to the latest master to include #525

daipom:
> Sorry, I forgot to point out a thing. Could you modify also fluent-package/templates/opt/fluent/share/td-agent.conf.tmpl?

By #525, we don't need to manage `td-agent.conf.tmpl` anymore.
I will check the package default behavior in some platforms after the CI completed

daipom:
Rebased to solve conflicts.

ashie (approved):

daipom:
I confirmed the actual logs of msi and rpm.
In addition, I added the tests for each platform to confirm the log files.

ashie (approved):

ashie:
Thanks!

daipom:
Thanks for your review!

PR #487, opened 2023-05-29
msi: migrate to /opt/fluent

kenhys:
Before:
 c:\opt\td-agent\

After:
 c:\opt\fluent\

Note that the following files are not migrated yet:

c:\opt\td-agent\etc\td-agent.conf
c:\opt\td-agent\etc\plugin
c:\opt\td-agent\td-agent.log
c:\opt\td-agent\var\log\td-agent

Thus, most of files are placed under c:\opt\fluent, but these
files are remained for keeping compatibility.

### DIFF
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -1555,7 +1555,7 @@ class WindowsPackageTask
          "-sreg",   # Suppress registry harvesting
          "-gg",                          # Generate guides
          "-cg", "ProjectDir",            # Component Group Name
-         "-dr", "PROJECTLOCATION",       # Root directory reference
+         "-dr", "FLUENTPROJECTLOCATION",       # Root directory reference
          "-var", "var.ProjectSourceDir", # Substitue File/@Source="SourceDir"
          "-t",   "exclude-files.xslt",   # XSLT for exclude files
          "-out", 'project-files.wxs')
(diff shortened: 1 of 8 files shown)


kenhys:
Result of changes:

opt-fluent

daipom:
Thanks for this fix!

> Note that the following files are not migrated yet:
> 
> c:\opt\td-agent\etc\td-agent.conf
> c:\opt\td-agent\etc\plugin
> c:\opt\td-agent\td-agent.log
> 
> Thus, most of files are placed under c:\opt\fluent, but these files are remained for keeping compatibility.

I'm concerned about `c:\opt\td-agent\var\log\td-agent\` too.
The default config uses this directory for buffer.
We need to figure out how to handle the compatibility of this buffer.

https://github.com/fluent/fluent-package-builder/blob/a0c9b5cd01bcd7c71d91cc8169c9b86931bf1952/fluent-package/templates/opt/fluent/share/td-agent.conf.tmpl#L19-L32

kenhys:
> I'm concerned about c:\opt\td-agent\var\log\td-agent\ too.
The default config uses this directory for buffer.

In this PR, fluentd.bat still point path to `c:\opt\td-agent\var\log`  via `TD_AGENT_TOPDIR`  environment variable.

daipom:
> > I'm concerned about c:\opt\td-agent\var\log\td-agent\ too.
> > The default config uses this directory for buffer.
> 
> In this PR, fluentd.bat still point path to `c:\opt\td-agent\var\log` via `TD_AGENT_TOPDIR` environment variable.

Understood! (We will handle the compatibility of these files in future PRs.)

> c:\opt\td-agent\etc\td-agent.conf
> c:\opt\td-agent\etc\plugin
> c:\opt\td-agent\td-agent.log
> c:\opt\td-agent\var\log\td-agent

ashie:
Thanks!

PR #489, opened 2023-05-30
migrate configuration files

kenhys:
In the previous versions, there are some configuration files and
log files are still kept under old path for compatibility.
It is better migrating to /opt/fluent completely, but we
need to consider upgrading from v4, too.

During upgrading from v4, seamless migration can be done by using
environment variable to specify the actual path because there is
no need to embed specific path to .service itself.
It aims not to be confused which configuration file should be applied.

Before:
 * Use old path for compatibility.
   * /etc/td-agent/td-agent.conf
   * /etc/td-agent/plugin
   * /var/log/td-agent/

After:

 * Use /opt/fluent as possible.
 * Keep /opt/td-agent as possible (only upgrading)
   * Moving content under /etc/td-agent to /etc/fluent/
   * Moving content under /var/log/td-agent to /var/log/fluent
   * Making symlink from /opt/td-agent /opt/fluent
   * Making symlink from /var/log/td-agent to /var/log/fluent
 * Note that /etc/logrotate.d/td-agent is not changed

PR #475, opened 2023-05-22
migrate to new user/group name

kenhys:
for deb, according Debian policy, dynamically allocated name should be prefixed
with an underscore.
(It may be not applicable for group name, but there is an prefixed
group name example e.g. _ssh/_cvsadmin)

See
https://www.debian.org/doc/debian-policy/ch-opersys.html#introduction

Note that adduser doesn't accept _name prefix by default. To avoid
it, we need to specify --force-badname explicitly

  * migration from to td-agent to _fluentd
    * service, hook script, tmpfiles.d and so on.
    * cleanup td-agent statoverride

For rpm, no prefixed underscore for user/group name.

  * migration from td-agent to fluentd
    * service, hook script, tmpfiles.d, permission and so on.
    * cleanup user/group on removing package

PR #493, opened 2023-06-08
dmg: migrate changed path

kenhys:
This is follow-up #489. it should be rebased with #475, #489.

Before:

* Use old path for compatibility.
  * /etc/td-agent/td-agent.conf
  * /var/log/td-agent/

After:

* Use /etc/fluent and /var/log/fluent by default.
* Keep /etc/td-agent, /var/log/td-agent as possible (only upgrading)
  * Moving content under /etc/td-agent to /etc/fluent/
  * Moving content under /var/log/td-agent to /var/log/fluent
  * Making symlink from /var/log/td-agent /var/log/fluent
  * Making symlink from /etc/td-agent to /etc/fluent

### DIFF
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -477,9 +477,11 @@ class BuildTask
 
       desc "Create td-agent configuration files from template"
       task :td_agent_config do
-        configs = [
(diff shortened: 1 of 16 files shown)


kenhys:
NOTE: just

PR #494, opened 2023-06-08
msi: migration to c:\opt\fluent

kenhys:
msi: migrate to c:\opt\fluent

Before:

* Configuration c:\opt\td-agent\etc\td-agent\td-agent.conf
* Log files: c:\opt\td-agent

After:

* Move configuration to c:\opt\fluent\etc\fluent\td-agent.conf
* Create symbolic link to c:\opt\fluent\etc\fluent from c:\opt\td-agent\etc\td-agent.
* Overwrite c:\opt\fluent\etc\fluentd.conf with td-agent.conf (when upgrade)
* Migrate to FLUENT_PACKAGE_TOPDIR environment variable.
* Keep old logs as is:

### DIFF
--- a/fluent-package/msi/assets/fluent-package-post-migration.bat
+++ b/fluent-package/msi/assets/fluent-package-post-migration.bat
@@ -1,0 +1,51 @@
+@echo on
+title Fluent-package post migration script
+
+if exist "%~dp0..\..\td-agent" (
+  setlocal enabledelayedexpansion
+  sc query state=inactive | findstr fluentdwinsvc
+  if !ERRORLEVEL! == 1 (
+    @rem NOTE: If the service is not stopped here, it is a BUG.
+    @rem This state can cause fatal problems with the installation, so we should cancel this process.
+    @rem (If trying to stop it by the command such as `$ net stop fluentdwinsvc`, it can fail as "The service is starting or stopping.  Please try again later.".)
+    echo "WARN: Active fluentdwinsvc service is detected. This is unexpected. Migration process has been canceled."
+    exit /b 1
+  )
+  endlocal
+
+  if not exist "%~dp0..\etc\fluent" (
+    echo "Create c:\opt\fluent\etc\fluent directory"
+    mkdir "%~dp0..\etc\fluent"
+  )
+
+  if exist "%~dp0..\..\td-agent\etc\td-agent\td-agent.conf" (
+    echo "Migrate c:\opt\td-agent\etc\td-agent.conf"
+    move /Y "%~dp0..\..\td-agent\etc\td-agent\td-agent.conf" "%~dp0..\etc\fluent\td-agent.conf"
+    echo "Migrate td-agent.conf to c:\opt\fluent\etc\fluent\fluentd.conf"
+    copy /Y "%~dp0..\etc\fluent\td-agent.conf" "%~dp0..\etc\fluent\fluentd.conf"
+  )
+
+  @rem NOTE: do not migrate log files not to lose log accidentally
+
+  echo "Migrate c:\opt\td-agent\etc\td-agent\* files"
+  for %%f in (%~dp0..\..\td-agent\etc\td-agent\*) do (
+    move /Y %%f "%~dp0..\etc\fluent\"
+  )
+  if not exist "%~dp0..\etc\fluent\plugin" (
+    mkdir "%~dp0..\etc\fluent\plugin"
+  )
+  echo "Migrate c:\opt\td-agent\etc\plugin directory"
+  for /d %%d in (%~dp0..\..\td-agent\etc\plugin\*) do (
+    move /Y %%d "%~dp0..\etc\fluent\plugin\"
+  )
+
+  echo "Ensure remaining files under td-agent"
+  tree %~dp0..\..\td-agent
+
+  @rem create symbolic link to c:\opt\fluent (we can't use hardlink for directory)
+  if exist "%~dp0..\..\td-agent\etc\td-agent" (
+    rmdir /S /Q "%~dp0..\..\td-agent\etc\td-agent"
+  )
+  echo "Create symlink c:\opt\td-agent\etc\td-agent to c:\opt\fluent\etc\fluent"
+  mklink /D %~dp0..\..\td-agent\etc\td-agent %~dp0..\..\fluent\etc\fluent
+)
--- a/fluent-package/msi/assets/fluent-package-post-migration.bat
+++ b/fluent-package/msi/assets/fluent-package-post-migration.bat
@@ -40,7 +40,7 @@ if exist "%~dp0..\..\td-agent" (
   )
 
   echo "Ensure remaining files under td-agent"
-  tree %~dp0..\..\td-agent
+  tree /F %~dp0..\..\td-agent
 
   @rem create symbolic link to c:\opt\fluent (we can't use hardlink for directory)
   if exist "%~dp0..\..\td-agent\etc\td-agent" (
--- a/fluent-package/msi/assets/fluentd.bat
+++ b/fluent-package/msi/assets/fluentd.bat
@@ -7,7 +7,7 @@ if "%~nx0" == "td-agent.bat" (
 
 @rem Convert path separator from backslash to forwardslash
 setlocal enabledelayedexpansion
-set FLUENT_PACKAGE_TOPDIR=FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!
+set FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!
 
 set PATH=%FLUENT_PACKAGE_TOPDIR%bin;%PATH%
 set PATH=%FLUENT_PACKAGE_TOPDIR%;%PATH%
--- a/fluent-package/msi/assets/fluentd.bat
+++ b/fluent-package/msi/assets/fluentd.bat
@@ -1,22 +1,19 @@
 @echo off
 if "%~nx0" == "td-agent.bat" (
   set FLUENT_PACKAGE_TOPDIR=%~dp0..\
-  set TD_AGENT_TOPDIR=%~dp0..\..\td-agent
 ) else (
   set FLUENT_PACKAGE_TOPDIR=%~dp0
-  set TD_AGENT_TOPDIR=%~dp0..\td-agent
 )
 
 @rem Convert path separator from backslash to forwardslash
 setlocal enabledelayedexpansion
 set FLUENT_PACKAGE_TOPDIR="FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!"
-set TD_AGENT_TOPDIR="TD_AGENT_TOPDIR=!TD_AGENT_TOPDIR:\=/!"
 endlocal
 
 set PATH=%FLUENT_PACKAGE_TOPDIR%bin;%PATH%
 set PATH=%FLUENT_PACKAGE_TOPDIR%;%PATH%
-set FLUENT_CONF=%TD_AGENT_TOPDIR%/etc/td-agent/td-agent.conf
-set FLUENT_PLUGIN=%TD_AGENT_TOPDIR%/etc/td-agent/plugin
+set FLUENT_CONF=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/fluentd.conf
+set FLUENT_PLUGIN=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/plugin
 set FLUENT_PACKAGE_VERSION=%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb
 for %%p in (%*) do (
     if "%%p"=="--version" (
--- a/fluent-package/msi/assets/fluentd.bat
+++ b/fluent-package/msi/assets/fluentd.bat
@@ -6,17 +6,24 @@ if "%~nx0" == "td-agent.bat" (
   set FLUENT_PACKAGE_TOPDIR=%~dp0
   set TD_AGENT_TOPDIR=%~dp0..\td-agent
 )
+
+@rem Convert path separator from backslash to forwardslash
+setlocal enabledelayedexpansion
+set FLUENT_PACKAGE_TOPDIR="FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!"
+set TD_AGENT_TOPDIR="TD_AGENT_TOPDIR=!TD_AGENT_TOPDIR:\=/!"
+endlocal
+
 set PATH=%FLUENT_PACKAGE_TOPDIR%bin;%PATH%
 set PATH=%FLUENT_PACKAGE_TOPDIR%;%PATH%
-set FLUENT_CONF=%TD_AGENT_TOPDIR%\etc\td-agent\td-agent.conf
-set FLUENT_PLUGIN=%TD_AGENT_TOPDIR%\etc\td-agent\plugin
-set FLUENT_PACKAGE_VERSION=%FLUENT_PACKAGE_TOPDIR%\bin\fluent-package-version.rb
+set FLUENT_CONF=%TD_AGENT_TOPDIR%/etc/td-agent/td-agent.conf
+set FLUENT_PLUGIN=%TD_AGENT_TOPDIR%/etc/td-agent/plugin
+set FLUENT_PACKAGE_VERSION=%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb
 for %%p in (%*) do (
     if "%%p"=="--version" (
         ruby "%FLUENT_PACKAGE_VERSION%"
         goto last
     )
 )
-"%FLUENT_PACKAGE_TOPDIR%\bin\fluentd" %*
+"%FLUENT_PACKAGE_TOPDIR%/bin/fluentd" %*
 
 :last
--- a/fluent-package/msi/assets/fluentd.bat
+++ b/fluent-package/msi/assets/fluentd.bat
@@ -7,8 +7,7 @@ if "%~nx0" == "td-agent.bat" (
 
 @rem Convert path separator from backslash to forwardslash
 setlocal enabledelayedexpansion
-set FLUENT_PACKAGE_TOPDIR="FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!"
-endlocal
+set FLUENT_PACKAGE_TOPDIR=FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!
 
 set PATH=%FLUENT_PACKAGE_TOPDIR%bin;%PATH%
 set PATH=%FLUENT_PACKAGE_TOPDIR%;%PATH%
@@ -22,5 +21,6 @@ for %%p in (%*) do (
     )
 )
 "%FLUENT_PACKAGE_TOPDIR%/bin/fluentd" %*
+endlocal
 
 :last
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -546,6 +546,7 @@ class BuildTask
         ensure_directory(staging_bindir)
         cp("msi/assets/#{PACKAGE_NAME}-prompt.bat", td_agent_staging_dir)
         cp("msi/assets/#{PACKAGE_NAME}-post-install.bat", staging_bindir)
+        cp("msi/assets/#{PACKAGE_NAME}-post-migration.bat", staging_bindir)
         cp("msi/assets/#{SERVICE_NAME}.bat", td_agent_staging_dir)
         cp("msi/assets/fluent-gem.bat", td_agent_staging_dir)
         cp("msi/assets/#{PACKAGE_NAME}-version.rb", staging_bindir)
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -477,11 +477,7 @@ class BuildTask
 
       desc "Create td-agent configuration files from template"
       task :td_agent_config do
-        configs = if windows?
-                    ["etc/#{COMPAT_PACKAGE_DIR}/#{COMPAT_SERVICE_NAME}.conf"]
-                  else
-                    ["etc/#{PACKAGE_DIR}/#{SERVICE_NAME}.conf"]
-                  end
+        configs = ["etc/#{PACKAGE_DIR}/#{SERVICE_NAME}.conf"]
         configs.concat([
           "etc/logrotate.d/#{SERVICE_NAME}",
           "opt/#{PACKAGE_DIR}/share/#{COMPAT_SERVICE_NAME}-ruby.conf",
@@ -489,8 +485,7 @@ class BuildTask
         ]) unless windows? || macos?
         configs.each do |config|
           src = template_path(config)
-          if config == "etc/#{PACKAGE_DIR}/#{SERVICE_NAME}.conf" or
-            config == "etc/#{COMPAT_PACKAGE_DIR}/#{COMPAT_SERVICE_NAME}.conf"
+          if config == "etc/#{PACKAGE_DIR}/#{SERVICE_NAME}.conf"
             src = template_path("opt/#{PACKAGE_DIR}/share/#{COMPAT_SERVICE_NAME}.conf.tmpl")
           end
           dest = File.join(STAGING_DIR, config)
--- a/fluent-package/msi/assets/fluent-package-post-install.bat
+++ b/fluent-package/msi/assets/fluent-package-post-install.bat
@@ -1,5 +1,5 @@
 @echo off
-title Td-agent post install script
+title Fluent-package post install script
 if not "%~dp0" == "C:\opt\fluent\bin\" (
    "%~dp0gem" pristine --only-executables --all
 )
--- a/fluent-package/msi/exclude-files.xslt
+++ b/fluent-package/msi/exclude-files.xslt
@@ -20,7 +20,7 @@
     <!-- Exclude patterns -->
     <xsl:key
         name="conf-search"
-        match="wix:Component[wix:File/@Source = '$(var.ProjectSourceDir)\etc\td-agent\td-agent.conf']"
+        match="wix:Component[wix:File/@Source = '$(var.ProjectSourceDir)\etc\fluent\fluentd.conf']"
         use="@Id" />
     <xsl:template match="*[self::wix:Component or self::wix:ComponentRef][key('conf-search', @Id)]" />
     <xsl:key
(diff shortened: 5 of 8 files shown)


kenhys:
[debug-msi.log](https://github.com/fluent/fluent-package-builder/files/11785407/debug-msi.log)

InstallExecuteSequence doesn't work as expected :thinking:

kenhys:
newly install works as expected, not for upgrade.
[pr494-new-install.log](https://github.com/fluent/fluent-package-builder/files/11785719/pr494-new-install.log)
[pr494-upgrade.log](https://github.com/fluent/fluent-package-builder/files/11785721/pr494-upgrade.log)

kenhys:
Unexpectedly action was skipped 

```
MSI (s) (50:1C) [16:52:42:112]: Note: 1: 2265 2:  3: -2147287035 
MSI (s) (50:1C) [16:52:42:112]: Server not locked: locking for product {4FEF158E-E67C-4D05-B7E2-55DA1B33AF9F}
Action ended 16:52:42: InstallInitialize. Return value 1.
MSI (s) (50:1C) [16:52:42:125]: Skipping action: SetDeleteCompatTdAgentBat (condition is false)
MSI (s) (50:1C) [16:52:42:125]: Skipping action: DeleteCompatTdAgentBat (condition is false)
MSI (s) (50:1C) [16:52:42:125]: Skipping action: SetDeleteCompatTdAgentGemBat (condition is false)
MSI (s) (50:1C) [16:52:42:125]: Skipping action: DeleteCompatTdAgentGemBat (condition is false)
MSI (s) (50:1C) [16:52:42:125]: Doing action: ProcessComponents
MSI (s) (50:1C) [16:52:42:125]: Note: 1: 2205 2:  3: ActionText 
Action 16:52:42: ProcessComponents. Updating component registration
Action start 16:52:42: ProcessComponents.
```

kenhys:
PostInstall action is called.

```
MSI (s) (50:1C) [16:53:06:342]: Executing op: ActionStart(Name=PostInstall,,)
Action 16:53:06: PostInstall. 
MSI (s) (50:1C) [16:53:06:344]: Executing op: CustomActionSchedule(Action=PostInstall,ActionType=3073,Source=BinaryData,Target=WixQuietExec64,CustomActionData="c:\opt\fluent\bin\fluent-package-post-install.bat")
MSI (s) (50:B4) [16:53:06:352]: Invoking remote custom action. DLL: C:\WINDOWS\Installer\MSICEAC.tmp, Entrypoint: WixQuietExec64
WixQuietExec64:  Restoring gems to pristine condition...
WixQuietExec64:  Skipped error_highlight-0.5.1, it is a default gem
```

kenhys:
postinstall's `if exist "%~dp0..\..\td-agent" (` condition may not be evaluated expectedly :thinking:

kenhys:
check error log:
[upgrade.take1.rollback.zip](https://github.com/fluent/fluent-package-builder/files/11796121/upgrade.take1.rollback.zip)

kenhys:
> check error log: [upgrade.take1.rollback.zip](https://github.com/fluent/fluent-package-builder/files/11796121/upgrade.take1.rollback.zip)

Should align not expected WixQuietExec64 execution order.

kenhys:
[upgrade-echo-on.u8.log](https://github.com/fluent/fluent-package-builder/files/11815470/upgrade-echo-on.u8.log)
tracked via @echo on, but it seems that the second `if exist ...` sequence is not executed.

kenhys:
It is out of scope, I've found that post-install was always called because of a mismatch...

```
WixQuietExec64:  C:\Windows\SysWOW64>if not "c:\opt\fluent\bin\" == "C:\opt\fluent\bin\" ("c:\opt\fluent\bin\gem" pristine --only-executables --all ) 
WixQuietExec64:  Restoring gems to pristine condition...
```

kenhys:
When executing extracted bat file content to simplified msi, it will be kicked. (even though weird mojibake errror)

kenhys:
with 7fb0059, WixQuietExec64 failure was fixed, but it works still unexpectdly, checking why...

kenhys:
c:\opt\td-agent was linked to c:\opt\fluent.

image

kenhys:
image

c:\opt\fluent\var\log\td-agent was linked to c:\opt\fluent\var\log\fluent.

daipom:
Thanks for this fix!
I will review this early next week.

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,46 @@
...
+  echo "Migrate c:\opt\td-agent\etc\td-agent\* files"
+  for %%f in (%~dp0..\..\td-agent\etc\td-agent\*) do (
+     move /Y %%f "%~dp0..\etc\fluent\"
+  )
+  echo "migrate c:\opt\td-agent\etc\plugin directory"
+  for /d %%d in (%~dp0..\..\td-agent\etc\plugin\*) do (
+     move /Y %%d "%~dp0..\etc\fluent\plugin\"
+  )
+  sc query state=inactive | findstr fluentdwinsvc
+  if !ERRORLEVEL! == 1 (
```
daipom, line 30: Is this typo?

```suggestion
  if %ERRORLEVEL% == 1 (
```

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,46 @@
...
+    echo "Active fluentdwinsvc service is detected"
+    sc stop fluentdwinsvc
+    timeout /t 5
```
daipom, line 33: I think we can use `net stop` to stop the service synchronously.
(`net` command would be very famous and commonly used command, but I can't find the official document of this command...)

```suggestion
    net stop fluentdwinsvc
```

fluent-package/msi/assets/fluent-package-post-migration.bat
daipom, line 30: Shouldn't this be the first process?
(right after the `if exist "%~dp0..\..\td-agent"`)

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,46 @@
...
+  )
+  echo "Ensure remaining files under td-agent"
+  tree %~dp0..\..\td-agent
+
+  @rem ensure to remove remained c:\opt\td-agent directory
+  rmdir /S /Q %~dp0..\..\td-agent
```
daipom, line 39: We may need to consider the possibility of renaming it instead of deleting it.

daipom (commented):
I commented on some points of concern.
I'm going to look at it in more detail, checking the actual behavior.

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: This is intentional, but I forgot to enable delayedexpansion.

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: Thank!

fluent-package/msi/assets/fluent-package-post-migration.bat
daipom: Oh! delayedexpansion! I see!

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,46 @@
...
+
+  @rem create symbolic link to c:\opt\fluent (we can't use hardlink for directory)
+  echo "Create symlink c:\opt\td-agent to c:\opt\fluent"
+  mklink /D %~dp0..\..\td-agent %~dp0..\..\fluent
+  echo "Create symlink c:\opt\fluent\var\log\td-agent to c:\opt\fluent\var\log\fluent"
+  mklink /D %~dp0..\..\fluent\var\log\td-agent %~dp0..\..\fluent\var\log\fluent
```
daipom, line 45: We also need the link `/opt/fluent/etc/td-agent` -> `/opt/fluent/etc/fluent` so that we can access the old conf file by `/opt/td-agent/etc/td-agent/td-agent.conf`.

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,46 @@
...
+  if not exist "%~dp0..\var\log\fluent" (
+     mkdir "%~dp0..\var\log\fluent"
+  )
+  echo "Migrate c:\opt\td-agent\var\log\td-agent\* files"
+  for %%f in (%~dp0..\..\td-agent\var\log\td-agent\*) do (
+     move /Y %%f "%~dp0..\var\log\fluent\"
+  )
+  echo "Migrate c:\opt\td-agent\var\log\td-agent\* directory (e.g buffer)"
+  for /d %%d in (%~dp0..\..\td-agent\var\log\td-agent\*) do (
+     move /Y %%d "%~dp0..\var\log\fluent\"
```
daipom, line 19: > Move log files to c:\opt\fluent\var\log\fluent.

I think this is the logic.
This logic certainly moves `var\log\` files, but they are actually the buffer files only.
Currently, log files are directly under `C:\opt\td-agent\`, and they are lost!
We should consider the log files migration.

* `C:\opt\td-agent\td-agent-supervisor-0.log`
* `C:\opt\td-agent\td-agent-0.log`

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: It seems better to change. thanks.

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: Good catch.

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: Thanks, right thing to do so.

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: Thanks. I'll change it.

daipom (commented):
> Move configuration files to c:\opt\fluent\etc\fluent\td-agent.conf

I have a question about this specification.
With only this, Fluentd will start with the default setting `c:\opt\fluent\etc\fluent\fluentd.conf` after the update.

In the deb/rpm package, the path to the configuration file is changed by an environment variable.
That is difficult for Windows, so I guess that we put that on hold in this PR, right?

I think we should consider overwriting the new default config file with the old file. 
I think we could consider it in common with msi, deb and rpm later.

kenhys:
also updated commit log.

daipom:
If the old td-agent service is running, this pop-up appears.
(Maybe it's been that way before. I'll check it.)

2023-06-27_13h33_46

daipom:
> If the old td-agent service is running, this pop-up appears. (Maybe it's been that way before. I'll check it.)
> 
> 2023-06-27_13h33_46

I confirmed it's been that way before. (This message appears though we don't need a reboot)
We should ignore this for now.

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,66 @@
...
+     move /Y %%f "%~dp0..\etc\fluent\"
+  )
+  echo "migrate c:\opt\td-agent\etc\plugin directory"
+  for /d %%d in (%~dp0..\..\td-agent\etc\plugin\*) do (
+     move /Y %%d "%~dp0..\etc\fluent\plugin\"
+  )
+  echo "Migrate c:\opt\td-agent\*.log files"
+  for %%f in (%~dp0..\..\td-agent\*.log) do (
+     move /Y %%f "%~dp0..\fluent\"
+  )
```
daipom, line 42: Somehow, this doesn't seem to work.
I confirmed the package behavior, but I couldn't find the old log files.

I'm reconsidering log file migration on Windows.
On Windows, even if the log file is moved, Fluentd will output the log to new files.
So, we may not need to move the log file.
(It remains in the backup-ed directory `td-agent.old`.)

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,66 @@
...
+  @rem check whether directory is empty or not.
+  dir /a:-d %~dp0..\..\td-agent
+  if !ERRORLEVEL! == 0 (
+    echo "Deleting empty c:\opt\td-agent"
+    rmdir /S /Q %~dp0..\..\td-agent
+  ) else {
+    echo "Usually empty directory is expected, If not, preserving %~dp0..\..\td-agent as a archive"
+    move /Y %~dp0..\..\td-agent %~dp0..\..\td-agent.old
+  )
+  endlocal
```
daipom, line 57: Somehow, this doesn't seem to work.
I'm investigating it now.

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,66 @@
...
+  tree %~dp0..\..\td-agent
+
+  @rem ensure to remove remained c:\opt\td-agent directory
+  setlocal enabledelayedexpansion
+  @rem check whether directory is empty or not.
+  dir /a:-d %~dp0..\..\td-agent
+  if !ERRORLEVEL! == 0 (
+    echo "Deleting empty c:\opt\td-agent"
+    rmdir /S /Q %~dp0..\..\td-agent
+  ) else {
```
daipom, line 53:
```suggestion
  ) else (
```

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,66 @@
...
+     move /Y %%f "%~dp0..\fluent\"
+  )
+  echo "Ensure remaining files under td-agent"
+  tree %~dp0..\..\td-agent
+
+  @rem ensure to remove remained c:\opt\td-agent directory
+  setlocal enabledelayedexpansion
+  @rem check whether directory is empty or not.
+  dir /a:-d %~dp0..\..\td-agent
+  if !ERRORLEVEL! == 0 (
```
daipom, line 50: This condition looks reversed.
`File Found` => `!ERRORLEVEL! == 0` => `preserve`

fluent-package/msi/assets/fluent-package-post-migration.bat
daipom: I have commented on the cause below.
However, if we should keep the old log files in `td-agent.old` without moving them, we should always create this directory. (commented above)

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: Thanks, I've fixed.

kenhys: Removing logic was omitted.

kenhys:
Try to reduce a risk of losing files.

kenhys:
changing path separator issue may be solved with "v=!v:\=/!" expression.

kenhys:
Need to fix:

```
WixQuietExec64:  ボリューム シリアル番号は 7830-FC66 です
WixQuietExec64:  C:\OPT\TD-AGENT
WixQuietExec64:  └─etc
WixQuietExec64:      └─td-agent
WixQuietExec64:  "Create symlink c:\opt\fluent\var\log\td-agent to c:\opt\fluent\var\log\fluent"
WixQuietExec64:  指定されたパスが見つかりません。
WixQuietExec64:  "Create symlink c:\opt\td-agent\etc\td-agent to c:\opt\fluent\etc\fluent"
WixQuietExec64:  既に存在するファイルを作成することはできません。
WixQuietExec64:  Error 0x80070001: Command line returned an error.
WixQuietExec64:  Error 0x80070001: QuietExec64 Failed
WixQuietExec64:  Error 0x80070001: Failed in ExecCommon method
```

kenhys:
msi

daipom:
Thanks! I will review this again!

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,47 @@
...
+title Fluent-package post migration script
+
+if exist "%~dp0..\..\td-agent" (
+  setlocal enabledelayedexpansion
+  sc query state=inactive | findstr fluentdwinsvc
+  if !ERRORLEVEL! == 1 (
+    echo "Active fluentdwinsvc service is detected"
+    net stop fluentdwinsvc
+  )
+  endlocal
```
daipom, line 11: I am wondering if it is possible for the service to be active at this time.
Against my expectations, it is judged to be active in my tests.
And `net stop` often fails.

```
WixQuietExec64:  "Active fluentdwinsvc service is detected"
WixQuietExec64:  The service is starting or stopping.  Please try again later.
```

Sometimes the installation does not work and this may have something to do with it.
I will check it.

daipom: I found the cause. 
At `InstallFluentdWinSvc` action, the service is started because of the option `--reg-winsvc-auto-start`, which is added here.
I think this makes the behavior unstable.

https://github.com/fluent/fluent-package-builder/blob/f285807b93b03c8338dd0b58371c247e6fbfd045/fluent-package/msi/source.wxs#L179-L181

We should disable this auto-start since few users want this auto-start.

* No one would be happy to start Fluentd with the default settings.
* It opens some ports by default. It is undesirable from a security point of view.

How about adding fix like this?: https://github.com/daipom/fluent-package-builder/commit/10ae6750565df7b1f70e0098404c8336565ced0a

I'm checking the CI results on my fork with this fix now.

daipom: The CI on my fork succeeds!: https://github.com/daipom/fluent-package-builder/actions/runs/5504485369

daipom: I confirmed the package built in the CI works as expected in my environment.

kenhys: It looks better to me, then could you push a commit on this pull request?

ashie: In addition, please add description about removing `--reg-winsvc-auto-start` to the first comment.

daipom: @kenhys Sure!

@ashie To the first comment of this PR?

ashie: > To the first comment of this PR?

Yes, it's used as the commit message on squash merging.

daipom: > > To the first comment of this PR?
> 
> Yes, it's used as the commit message on squash merging.

Instead of adding the change to this PR, I have made another PR:

* #521

I want to rebase this branch on it after it is merged.

daipom: > It looks better to me, then could you push a commit on this pull request?

I have applied the changes except #521!

daipom:
I have rebased this to apply

* #521

fluent-package/msi/assets/fluent-package-post-migration.bat
daipom: Solved by #521.

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,51 @@
...
+  if not exist "%~dp0..\etc\fluent\plugin" (
+    mkdir "%~dp0..\etc\fluent\plugin"
+  )
+  echo "Migrate c:\opt\td-agent\etc\plugin directory"
+  for /d %%d in (%~dp0..\..\td-agent\etc\plugin\*) do (
+    move /Y %%d "%~dp0..\etc\fluent\plugin\"
+  )
+
+  echo "Ensure remaining files under td-agent"
+  tree %~dp0..\..\td-agent
```
daipom, line 43: Is this for debugging?
If we need file lists for debugging, then should we use `/F` option?

fluent-package/msi/assets/fluentd.bat
```diff
@@ -1,22 +1,26 @@
...
-  set TD_AGENT_TOPDIR=%~dp0..\..\td-agent
 ) else (
   set FLUENT_PACKAGE_TOPDIR=%~dp0
-  set TD_AGENT_TOPDIR=%~dp0..\td-agent
 )
+
+@rem Convert path separator from backslash to forwardslash
+setlocal enabledelayedexpansion
+set FLUENT_PACKAGE_TOPDIR="FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!"
+endlocal
```
daipom, line 11: Somehow, this does not appear to work for

* https://github.com/fluent/fluentd/issues/1138

I tested the package with the following config, but backslash was still used and could not load the existing buffer files on restart.

Maybe we should handle this separately from this PR.

```xml
<source>
  @type sample
  tag test
</source>

<match test.**>
  @type stdout
  <buffer>
    @type file
    path "#{ENV['FLUENT_PACKAGE_TOPDIR']}/var/log/fluentd/buffer/stdout"
    flush_at_shutdown false
    flush_interval 10
  </buffer>
</match>
```

log

```
[info]: using configuration file: <ROOT>
  <source>
    @type sample
    tag "test"
  </source>
  <match test.**>
    @type stdout
    <buffer>
      @type "file"
      path "C:\\opt\\fluent\\/var/log/fluentd/buffer/stdout"
      flush_at_shutdown false
      flush_interval 10
    </buffer>
  </match>
</ROOT>
```

daipom (commented):
Now the migration process works stably!

I have several concerns on a point other than the migration process itself.
I have commented them below.

fluent-package/msi/assets/fluent-package-post-migration.bat
```diff
@@ -0,0 +1,51 @@
...
+
+if exist "%~dp0..\..\td-agent" (
+  setlocal enabledelayedexpansion
+  sc query state=inactive | findstr fluentdwinsvc
+  if !ERRORLEVEL! == 1 (
+    @rem NOTE: If the service is not stopped here, it is a BUG.
+    @rem This state can cause fatal problems with the installation, so we should cancel this process.
+    @rem (If trying to stop it by the command such as `$ net stop fluentdwinsvc`, it can fail as "The service is starting or stopping.  Please try again later.".)
+    echo "WARN: Active fluentdwinsvc service is detected. This is unexpected. Migration process has been canceled."
+    exit /b 1
```
daipom, line 12: I checked this behavior by rewriting the codes.
When exiting ERRORLEVEL `1`, Rollback doesn't occur, and the installation is completed without running the migration process in this script.

I think this is acceptable for now.
Fixes to ensure Rollback should be done in future work.

ashie: I think so too.

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: Yes, it is for debugging use. it is good idea to apply /F

fluent-package/msi/assets/fluent-package-post-migration.bat
kenhys: Thanks for the confirmation. :+1:

fluent-package/msi/assets/fluentd.bat
kenhys: it will be fixed via ab9df59

kenhys: Here is the example to verify scope of enabledelayedexpansion:

```
set FLUENT_PACKAGE_TOPDIR=%~dp0..\
set TD_AGENT_TOPDIR=%~dp0..\td-agent
setlocal enabledelayedexpansion
echo "FLUENT_PACKAGE_TOPDIR:"
set FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!
set TD_AGENT_TOPDIR=!TD_AGENT_TOPDIR:\=/!
echo %TD_AGENT_TOPDIR%
echo %FLUENT_PACKAGE_TOPDIR%
endlocal

echo "OUTSIDE"
echo %TD_AGENT_TOPDIR%
echo %FLUENT_PACKAGE_TOPDIR%
```

expansion

daipom: Wow! I see! Thanks for the fix!

daipom: I will check it after the CI finished and the new package is built.

ashie:
Install test seems failed since ab9df591684fc63283da50df695f05f5ff0ac373
```
Installing C:\fluentd\fluent-package\msi\repositories\fluent-package-5.0.0-x64.msi 
Start-Service : Cannot find any service with service name 'fluentdwinsvc'. 
At C:\fluentd\fluent-package\msi\install-test.ps1:7 char:1
+ Start-Service fluentdwinsvc
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (fluentdwinsvc:String) [Start-Se  
   rvice], ServiceCommandException
    + FullyQualifiedErrorId : NoServiceFoundForGivenName,Microsoft.PowerShell.  
   Commands.StartServiceCommand
```

kenhys:
Redundant expression causes registration error.
It will be fixed via 40bc79d.

daipom:
Thanks! I will check the behavior of the package after CI finishes.

fluent-package/msi/assets/fluentd.bat
daipom: https://github.com/fluent/fluent-package-builder/pull/494#discussion_r1259005439 seems still not solved.
I checked the package, but the result is the same.

Sorry, there are some points I don't understand.

https://github.com/fluent/fluent-package-builder/pull/494#discussion_r1259179257 means the slash conversion is lost outside the scope?
If so, the environmental variable `ENV['FLUENT_PACKAGE_TOPDIR']` in the config should still contain backslashes?

I checked the previous environmental variable `ENV['TD_AGENT_TOPDIR']` in the config.
It returns `C:\\opt\\td-agent\\`.
I think it is strange...
Now it seems to me that no one sets this old variable anymore.
Why can Fluentd still access this variable...

I think it is correct that this old environment variable still works as is.
If not, then the old config using this variable can't work in v5.
But I don't understand how this still works now...

daipom: Summary

* `ENV['TD_AGENT_TOPDIR']` should work in the new package, but it is uncertain in the current implementation.
* The logic converting BachSlash in `ENV['FLUENT_PACKAGE_TOPDIR']` works for the environmental variables, but it would not be used from the configuration.

We should handle these points in other PRs.

daipom (approved):
LGTM.

We should handle the points I commented on https://github.com/fluent/fluent-package-builder/pull/494#discussion_r1263217312 in other PRs.

PR #495, opened 2023-06-09
deb: test with systemd

daipom:
This adds a test for deb packages to confirm the package works correctly with systemd.

There are still some points that need more refactoring, but it is better to have such a test as soon as possible.

The current `Vagrantfile` is just for manual operation and is not used recently, so I fixed it entirely for CI.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,49 +7,26 @@ VAGRANTFILE_API_VERSION = "2"
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   vms = [
     {
-      :id => "debian-buster-amd64",
-      :box => "debian/buster64",
+      :id => "debian-bullseye",
+      :box => "bento/debian-11",
     },
     {
-      :id => "ubuntu-16.04-x86_64",
-      :box => "ubuntu/xenial64",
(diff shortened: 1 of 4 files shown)


Vagrantfile
```diff
@@ -7,49 +7,26 @@ VAGRANTFILE_API_VERSION = "2"
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   vms = [
     {
-      :id => "debian-buster-amd64",
-      :box => "debian/buster64",
+      :id => "debian-bullseye",
+      :box => "bento/debian-11.3",
     },
```
kenhys, line 12: bento/debian-11.3 is a bit old point release, how about using bento/debian-11?
at least we can test with debian 11.6 point release.

kenhys: Except above issue, LGTM.

daipom: Thanks! I will check it!

daipom: Fixed!
Thanks!

daipom:
All green!

kenhys:
LGTM.

kenhys:
merged.

daipom:
Thanks!

PR #496, opened 2023-06-09
rpm: test with systemd

daipom:
The rpm version of #495

There is no `AmazonLinux2023` vagrant image, so I omit it for now.


kenhys (approved):
LGTM.

daipom:
Thanks!

PR #498, opened 2023-06-14
apt: apply --no-pager to disable pager on tty

kenhys:
When manually testing with script, script will be interrupted with pager. This commit explicitly disable such feature.


daipom (approved):
LGTM.
I wasn't aware of this possibility. Thanks for fixing this!

PR #499, opened 2023-06-14
apt: enable systemd service test again

kenhys:
follow-up of #489

PR #500, opened 2023-06-14
Use new path for fluent-package

kenhys:
No need to use old path for fluent-package.

### DIFF
--- a/fluent-package/apt/systemd-test/update-from-v4.sh
+++ b/fluent-package/apt/systemd-test/update-from-v4.sh
@@ -11,14 +11,14 @@ sudo apt install -V -y ${apt_source_package} ca-certificates
 sudo apt update
 sudo apt install -V -y td-agent=4.5.0-1
 
-systemctl status --no-pager td-agent
+systemctl status --wait --no-pager td-agent
 
 # Install the current
 sudo apt install -V -y \
     /vagrant/${distribution}/pool/${code_name}/${channel}/*/*/fluent-package_*_${architecture}.deb
 
 # Test: service status
-systemctl status --no-pager fluentd
+systemctl status --wait --no-pager fluentd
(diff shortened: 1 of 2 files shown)


kenhys:
Need to rebase with #499

kenhys:
failure on focal and jammy. checking....

daipom:
It sometimes fails in other PRs as well. (#501)

Is it possible that https://github.com/fluent/fluent-package-builder/pull/489#issuecomment-1587075397 problem still remains?

kenhys:
> Is it possible that [#489 (comment)](https://github.com/fluent/fluent-package-builder/pull/489#issuecomment-1587075397) problem still remains?

There is a possibility of it. checking...

kenhys:
Try to run the failed job again.

kenhys:
I'm doubt that this fragile behavior comes from running systemctl immediately. (The state of current status is not synced yet) 
maybe waiting appropriate time will work. :thinking:

daipom:
> I'm doubt that this fragile behavior comes from running systemctl immediately. (The state of current status is not synced yet) maybe waiting appropriate time will work. thinking

I wonder if this is the real cause of this problem.

I have tested the following 2 patterns on my local.

* Add `sleep 3` before `systemctl status` check like this PR.
* Add `--wait` option to `systemctl status` check.

However, both patterns fail about 1 in 3 times.

It may be a good idea to add `sleep` or `--wait` option just in case, but perhaps it has nothing to do with this problem.

daipom (approved):
I reran tests several times and confirmed they can still fail.
However, I think it has nothing to do with this PR.

This PR looks good to me.

kenhys:
> However, both patterns fail about 1 in 3 times.

Thank you for checking.
I'll investigate it again.

Anyway, using `--wait` option is better, I'll fix it.

kenhys:
changed to use `--wait` option to sync.

PR #502, opened 2023-06-19
deb: tweak lintian warnings

kenhys:
* Drop unknown tag
   * Ignore E: unstripped-binary-or-object
    * Fix W: transitional-package-not-oldlibs-optional
    * Suppress W: unusual-interpreter
    * Suppress W: debian-rules-should-not-use-custom-compression-settings
    * Suppress E: custom-library-search-path
    * Suppress W: custom-compression-in-debian-rules


fluent-package/debian/fluent-package.lintian-overrides
```diff
@@ -31,3 +26,5 @@ fluent-package: binary-or-shlib-defines-rpath opt/fluent/lib/ruby/gems/*/gems/no
 fluent-package: embedded-library opt/fluent/lib/ruby/gems/*/extensions/x86_64-linux/*/nokogiri-*/nokogiri/nokogiri.so: libxml2
 fluent-package: embedded-library opt/fluent/lib/ruby/gems/*/gems/nokogiri-*/ext/nokogiri/nokogiri.so: libxml2
 fluent-package: embedded-library opt/fluent/lib/ruby/gems/*/gems/nokogiri-*/lib/nokogiri/nokogiri.so: libxml2
+fluent-package: unstripped-binary-or-object opt/fluent/lib/ruby/gems/3.2.0/extensions/x86_64-linux/3.2.0/debug-1.7.1/debug/debug.so
+fluent-package: unusual-interpreter opt/fluent/lib/ruby/gems/3.2.0/gems/ffi-1.15.5/ext/ffi_c/libffi/generate-darwin-source-and-headers.py #!python
```
cosmo0920, line 30: nits: Shouldn't we use `*` for version numbers?

kenhys: Thanks, as you mentioned, it should use `*` for it.

fluent-package/debian/fluent-package.lintian-overrides
```diff
@@ -4,17 +4,12 @@ fluent-package: script-not-executable
 fluent-package: python-script-but-no-python-dep
```
daipom, line 4: Shouldn't we remove `python-script-but-no-python-dep` too?

```suggestion
```

daipom (commented):
Shouldn't we update the man of `td` in this PR?

https://github.com/fluent/fluent-package-builder/blob/7221f3ee05e0458b3962362fa68f40768a615823/fluent-package/debian/td.1

daipom (approved):
My concerns are the above two points.
Other than those points, this looks good to me!

kenhys:
> Shouldn't we update the man of `td` in this PR?
> 
> https://github.com/fluent/fluent-package-builder/blob/7221f3ee05e0458b3962362fa68f40768a615823/fluent-package/debian/td.1

Good catch!, it should be updated too.

fluent-package/debian/fluent-package.lintian-overrides
kenhys: Thanks, fixed now.

kenhys:
waiting CI again.

daipom (approved):
LGTM!

kenhys:
Thanks!

PR #504, opened 2023-06-26
deb: fix warnings

kenhys:
follow up of #502 

* Fix W: command-with-path-in-maintainer-script
  Changed not to depend on existence of actual path.
* Suppress W: maintainer-script-calls-systemctl
* Suppress no-manual-page for deprecated td-agent, td-agent-gem.
  * no-manual-page is unknown on focal, but as binary-without-manpage
    is deprecated, so we dare to use no-manual-page.
* Bump compatibility levels to v12 (same as fluent-package)
* Bump Standards-Version to 4.5.0.


kenhys:
TODO: fix old standard version.

kenhys:
Fixed. also updated description, too.

daipom (approved):
LGTM!

Note for me:

* The version of `debhelper-compat` and `Standards-Version` is determined based on the oldest supported distribution (focal currently).
  * `Standards-Version` is the version of `debian-policy` package. https://packages.ubuntu.com/search?keywords=debian-policy&searchon=names&suite=focal&section=all
* On focal, `binary-without-manpage` warning still occurs, but we can ignore it since it is not error level, and that tag is replaced with `no-manual-page` after focal.
* Regarding simplifying `unusual-interpreter`, there is a risk of not being able to detect unexpected files, but it seems better than having a complex file specification for now.

PR #505, opened 2023-06-26
Migrate path of logrotate.d configration

kenhys:
Before: /etc/logrotate.d/td-agent
      If td-agent user/group exists, renaming td-agent
      to fluent user/group then rewrite logrotate.d configuration.
      But, it may fail if user modify that configuration unexpected
      way.

    After: /etc/logrotate.d/fluentd

      If td-agent user/group exists (e.g upgrading from v4), create
      _fluent user/group as same uid/gid. If not, just create new _fluent
      user/group (no td-agent user/group).

      This approach make easy to migrate /etc/logrotate.d/td-agent
      configuration file without modification.

    Additionally, rewrite /etc/logrotate.d/td-agent path of pid
    because the pid file will be dynamically created/destroyed.

### DIFF
--- a/fluent-package/templates/etc/logrotate.d/fluentd
+++ b/fluent-package/templates/etc/logrotate.d/fluentd
@@ -1,4 +1,4 @@
-/var/log/<%= compat_package_dir %>/<%= service_name %>.log {
+/var/log/<%= package_dir %>/<%= service_name %>.log {
   daily
   rotate 30
   compress
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -483,7 +483,7 @@ class BuildTask
                     ["etc/#{PACKAGE_DIR}/#{SERVICE_NAME}.conf"]
                   end
         configs.concat([
-          "etc/logrotate.d/#{COMPAT_SERVICE_NAME}",
+          "etc/logrotate.d/#{SERVICE_NAME}",
           "opt/#{PACKAGE_DIR}/share/#{COMPAT_SERVICE_NAME}-ruby.conf",
           "opt/#{PACKAGE_DIR}/share/#{COMPAT_SERVICE_NAME}.conf.tmpl"
         ]) unless windows? || macos?
--- a/fluent-package/apt/install-test.sh
+++ b/fluent-package/apt/install-test.sh
@@ -80,19 +80,21 @@ if [ ! -h /etc/td-agent ]; then
     exit 1
 fi
 
+# Note: As td-agent and _fluentd use same UID/GID,
+# it is regarded as preceding name (td-agent)
 owner=$(stat --format "%U/%G" /etc/fluent)
-if [ "$owner" != "_fluentd/_fluentd" ]; then
-    echo "/etc/fluent must be owned by _fluentd/_fluentd"
+if [ "$owner" != "td-agent/td-agent" ]; then
+    echo "/etc/fluent must be owned by td-agent/td-agent"
     exit 1
 fi
 owner=$(stat --format "%U/%G" /var/log/fluent)
-if [ "$owner" != "_fluentd/_fluentd" ]; then
-    echo "/var/log/fluent must be owned by _fluentd/_fluentd"
+if [ "$owner" != "td-agent/td-agent" ]; then
+    echo "/var/log/fluent must be owned by td-agent/td-agent"
     exit 1
 fi
 owner=$(stat --format "%U/%G" /var/run/fluent)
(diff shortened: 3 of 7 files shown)


daipom (commented):
Thanks for this fix!
Indeed, we have to update the file `/etc/logrotate.d` for the new package.
Do we also need to update the USER and GROUP name to `_fluentd` for the deb package?

For migration, I have the following concerns.

* Shouldn't the old rotate config `/etc/logrotate.d/td-agent` work as is after the update?
  * Because we have the alias to access the old path: `/var/log/td-agent/td-agent.log`
* So, I think we should not make the new rotate config, and continue to use the old one as is.
  * Possibly, we need to migrate USER and GROUP settings, but the other migration process seems to be not needed. 

I confirmed the alias `/var/log/td-agent` work as expected for rotation:

`/etc/logrotate.d/td-agent`

```ini
/var/log/td-agent/td-agent.log {
  #daily
  size 1
  rotate 30
  compress
  delaycompress
  notifempty
  create 640 _fluentd _fluentd
  sharedscripts
  postrotate
    pid=/var/run/fluent/fluentd.pid
    if [ -s "$pid" ]
    then
      kill -USR1 "$(cat $pid)"
    fi
  endscript
}
```

log files:

```
/var/log/fluent/td-agent.log
```

Run rotate: `$ sudo logrotate -s state /etc/logrotate.d/td-agent`

rotated log files:

```
/var/log/fluent/td-agent.log
/var/log/fluent/td-agent.log.1
```

kenhys:
It may be a simple solution to do:
* Ship /etc/logrotate.d/flentd (it use new path)
  * thus, both of /etc/logrotate.d/td-agent and /etc/logrotate.d/fluentd is effective. 
* Replace user/group in /etc/logrotate.d/td-agent
*

kenhys:
Try 3rd approach whether it will work as expected.

kenhys:
3rd approach is not bad, but tend to be broken on some deb platform. :thinking: 
we need to tackle on another PR.

fluent-package/templates/etc/logrotate.d/fluentd
daipom, line 1: This is a problem.
Since the pid file is changed to `/var/run/fluent/fluentd.pid` from `/var/run/td-agent/td-agent.pid`, the `SIGUSR1` signal is not sent on the environment migrated from v4.

```
postrotate
    pid=/var/run/<%= package_dir %>/<%= service_name %>.pid
    if [ -s "$pid" ]
    then
      kill -USR1 "$(cat $pid)"
    fi
endscript
```

daipom (commented):
I confirmed this basically works as expected on Ubuntu Focal.

There was one problem about `SIGUSR1` signal on rotation.
I commented on it below.

fluent-package/templates/etc/logrotate.d/fluentd
daipom: Is it acceptable to create such a link as follows?

```
/var/run/td-agent/td-agent.pid => /var/run/fluent/fluentd.pid
```

kenhys: Thanks, I'll chech it.

kenhys: /var/run will be dynamically created, thus such a link may be a dead link.

kenhys: > /var/run` will be dynamically created, thus such a link may be a dead link.

As service specify Type=forking, using ExecStartPost for creating symlink is not appropriate (it will block)

kenhys: Rewriting path of pid seems simple solution.
`

kenhys:
Added fix for logrotate.d/td-agent

fluent-package/templates/etc/logrotate.d/fluentd
daipom: I see!

daipom:
Is it intentional that https://github.com/fluent/fluent-package-builder/pull/505/commits/43e9343c7cb0a58234f0c7684dd113742df824c3 was added into this PR?

kenhys:
> Is it intentional that [43e9343](https://github.com/fluent/fluent-package-builder/commit/43e9343c7cb0a58234f0c7684dd113742df824c3) was added into this PR?

It should be seperate PR.

daipom (approved):
I confirmed that the SIGUSR1 worked correctly on rotation!
LGTM.

kenhys:
Thanks!

PR #509, opened 2023-06-27
Update target distributions

kenhys:
Both of distributions are already removed from GitHub Actions.

* buster: EOL (2022-09-10), though LTS is still alive (2024-06-30)
* bionic: End of Standard Support (June 2023), EOL April 2028

See
    https://wiki.debian.org/DebianReleases
    https://wiki.ubuntu.com/Releases

### DIFF
--- a/fluent-package/convert-artifacts-layout.sh
+++ b/fluent-package/convert-artifacts-layout.sh
@@ -15,9 +15,9 @@ case $1 in
     apt|deb)
 	REPOSITORY_TYPE=apt
 	REPOSITORY_PATH=$TD_AGENT_DIR/$REPOSITORY_TYPE/repositories
-	for d in buster bullseye bionic focal jammy; do
+	for d in bullseye bookworm focal jammy; do
 	    case $d in
-		buster|bullseye)
+		bullseye|bookworm)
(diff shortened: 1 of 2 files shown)


kenhys:
As #510 was merged, rebased it.

fluent-package/convert-artifacts-layout.sh
```diff
@@ -15,17 +15,17 @@ case $1 in
     apt|deb)
 	REPOSITORY_TYPE=apt
 	REPOSITORY_PATH=$TD_AGENT_DIR/$REPOSITORY_TYPE/repositories
-	for d in buster bullseye bionic focal jammy; do
+	for d in bullseye bookworm focal jammy; do
 	    case $d in
-		buster|bullseye)
+		bullseye)
```
daipom, line 20: Although `bookworm` is added to the list, there is no case for `bookworm`.
Is this expected?
Is it because we don't support it yet?

kenhys: I've overlooked it, thanks.

kenhys:
I've fixed it.

daipom (approved):
LGTM!

PR #510, opened 2023-06-28
test:  disable check with td-agent alias status

kenhys:
td-agent (alias of fluentd service) should be enabled, but systemctl status td-agent may fail frequently with exit 3. It means inactive (dead) status.
Until finding correct solution to fix this issue, disable it for a while.


daipom (approved):
LGTM. Thanks!

kenhys:
As #504 was merged, rebased with it.

PR #507, opened 2023-06-26
Add new signing key

kenhys:
* signing key is generated with gpg --full-generate-key [1]
* fix missing transitional package
  * It seems that transitional package must be placed with
  fluent-package. This is a bug of convert-artifacts-layout.sh.

[1]
RSA4096bit, Fluentd developers (Fluent Package Official Signing Key) <fluentd@googlegroups.com>.
Note that v5.0.0 must be singed with old key, then for next update,
we can use signing with the new key which is bundled with
fluentd-apt-source package.

NOTE: It seems that transitional package must be placed with
fluent-package. Without it, it causes the following error:

  E: Failed to fetch
  http://.../5/debian/bullseye/pool/contrib/f/fluent-package/td-agent_5.0.0-1_all.deb

Closes: #506

PR #511, opened 2023-06-29
deb: don't terminate test case for jammy

kenhys:
This is follow-up #507. need to rebase it.

Ubuntu 22.04 was supported since 4.3.1-1, so
we can enable it.


daipom (approved):
LGTM!

kenhys:
conflicted with HEAD, so rebased again.

PR #513, opened 2023-07-05
Tweak utility script for manage repository

kenhys:
* Add missing target to manage fluentd-apt-source
* Bump dry-download version to 5
* Add missing -architectures="all" for publishing only _all.deb
* Update supported platform for dry-upload


daipom (approved):
LGTM. Thanks!

Note for me:

* Probably, we need to update `fluentd-apt-source` manually until now.
* The reason `aptly` needs specifying of `architectures` is probably because `fluentd-apt-source` is added. (It is `all` package)

PR #514, opened 2023-07-05
deb: refine fluentd-apt-source 

kenhys:
* Update description fluentd-apt-source
* Drop dependency to deprecated transitional apt-transport-https.
  https support has been moved into the apt package.

### DIFF
--- a/fluent-package/debian/control
+++ b/fluent-package/debian/control
@@ -25,9 +25,10 @@ Pre-Depends: adduser
 Depends:
   ${misc:Depends},
   ${shlibs:Depends},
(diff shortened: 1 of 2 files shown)


daipom (approved):
Thanks! LGTM.

It seems that we should fix this too. (We can do it in another PR)

https://github.com/fluent/fluent-package-builder/blob/7d125dd37ae798a7a5989c58f8b11e04aa852b27/fluent-package/debian/control#L20-L32

kenhys:
Only updated description in this PR.

Renaming package name will be applied in another PR.

daipom (approved):
LGTM. Thanks!

PR #519, opened 2023-07-10
Set system account (nologin) and home directory

kenhys:
In the previous commit, useradd didn't set login shell
and home directory correctly.


ashie (approved):

kenhys:
rebased with HEAD and fix a typo about user name.

PR #521, opened 2023-07-10
msi: Disable auto start after install

daipom:
Many users would be happier if it didn't start automatically:

* No one would be happy to start Fluentd with the default settings.
* It opens some ports by default. It is undesirable from a security point of view.

In addition, we should make sure that the service starts after all install actions are completed.
To achieve it, this is preferable for now.


daipom:
> In addition, we should make sure that the service starts after all install actions are completed.

We actually need this for

* #494

ashie (approved):
It'll be nice that the installer's exit dialog notice the user needs to start the service manually.
https://wixtoolset.org/docs/v3/wixui/wixui_customizations/#customizing-the-exitdlg
I don't require it in this PR though.

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #522, opened 2023-07-12
Update bundle gems

kenhys:
It aims to check dependency tree.

PR #523, opened 2023-07-14
Try to use Jemalloc 5.3.0

kenhys:
Just evaluate 5.3.0, will not be merged.

pushed on #522

### DIFF
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -668,6 +668,7 @@ class BuildTask
     end
 
     cd(source_dir) do
+      sh("echo", "je_cv_madv_free=no", {out: "config.cache"})
       sh("./configure", *configure_opts)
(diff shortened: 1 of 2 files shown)


ashie:
Thanks, I also want to evaluate it.
One more thing, please consider disabling `MADV_FREE` https://github.com/fluent/fluentd-docker-image/pull/350

ashie:
I want to apply it if we observe no noticeable difference in memory usage with the following procedure:
https://github.com/fluent/fluentd/issues/1657#issue-248177661

kenhys:
visualized the result.

jemalloc5

ashie:
BTW the config names `test-in.conf` and `test-out.conf` are inverted (comes from the first comment of the issue).

* test-in.conf: forwarder
* test-out.conf: aggregator

kenhys:
It seems that it is reasonable to stay with jemalloc 3.x in point view of performance for a while.
It does not mean that declining migration to jemalloc 5.x in the future.

PR #524, opened 2023-07-14
apt: Fix an unstable systemd test

ashie:
Add `--now` to `systemctl enable` in update-from-v4.sh to make sure to start the service before checking the status.

### DIFF
--- a/fluent-package/apt/systemd-test/update-from-v4.sh
+++ b/fluent-package/apt/systemd-test/update-from-v4.sh
@@ -31,16 +31,9 @@ systemctl status --wait --no-pager fluentd
 
 # Test: restoring td-agent service alias
 sudo systemctl unmask td-agent
-sudo systemctl enable fluentd
-#
-# FIXME:
-# td-agent (alias of fluentd service should be enabled,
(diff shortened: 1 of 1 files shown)


daipom:
Wow! Thanks! I will merge this after 

* #512

daipom:
Certainly, it looks like we need `--now` option.

Note: man systemctl

```
 Enable one or more units or unit instances. This will create a set of symlinks, as encoded in the [Install] sections of the indicated unit files. After the symlinks have been created, the system manager configuration is
           reloaded (in a way equivalent to daemon-reload), in order to ensure the changes are taken into account immediately. Note that this does not have the effect of also starting any of the units being enabled. If this is
           desired, combine this command with the --now switch, or invoke start with appropriate arguments later. Note that in case of unit instance enablement (i.e. enablement of units of the form foo@bar.service), symlinks named
           the same as instances are created in the unit configuration directory, however they point to the single template unit file they are instantiated from.
```

I wonder why it sometimes works as expected without this setting or without `systemctl start`...

daipom (approved):
Looks good to me though I have a question.

https://github.com/fluent/fluent-package-builder/pull/524#issuecomment-1636735783

PR #101, opened 2020-06-12
rpm: fix W: non-conffile-in-etc

kenhys:
template file should not put on /etc

PR #525, opened 2023-07-17
Stop maintaining multiple copies of the same conf

daipom:
Just refactoring. This does not change the result.


daipom:
We need to do some fixes following this.
I will do them after this PR.

* Update the Fluentd config for Windows.
* Update the following filepaths.
  * `/opt/fluent/share/td-agent.conf.tmpl`
    * I think the filename should be `fluentd.conf.default`.
  * `/opt/fluent/share/td-agent-ruby.conf`

I have one question.
Do we still need `/opt/fluent/share/td-agent.conf.tmpl`?

It is not used after the following PR:

* #344

but it is still better to have the default config under the `/opt/fluent/share/`, right?

daipom:
Note: related fixes in the past:

* #344
* #147
* #101
* #3

ashie:
> but it is still better to have the default config under the /opt/fluent/share/, right?

I think so.

daipom:
> > but it is still better to have the default config under the /opt/fluent/share/, right?
> 
> I think so.

Thanks!

kenhys (approved):
As the rest of the work will be done in another PR (https://github.com/fluent/fluent-package-builder/pull/525#issuecomment-1638221669),
it seems a reasonable fix.

daipom:
Thanks for your review!

daipom:
Done all TODO!

https://github.com/fluent/fluent-package-builder/pull/525#issuecomment-1638221669

PR #526, opened 2023-07-17
deb: Stop using `--wait` option of systemctl

daipom:
We misunderstood the use of this option.

By default, `systemctl status` waits until the unit's start-up is completed (without `--no-block` option).
The option `--wait` is for blocking `systemctl start foobar` until the service `foobar` is stopped.

### DIFF
--- a/fluent-package/apt/systemd-test/update-from-v4.sh
+++ b/fluent-package/apt/systemd-test/update-from-v4.sh
@@ -17,22 +17,22 @@ sudo apt clean all
 sudo apt update
 sudo apt install -V -y td-agent=4.5.0-1
 
-systemctl status --wait --no-pager td-agent
+systemctl status --no-pager td-agent
 
 # Install the current
 sudo apt install -V -y \
     /vagrant/${distribution}/pool/${code_name}/${channel}/*/*/fluent-package_*_${architecture}.deb
 
 # Test: service status
-systemctl status --wait --no-pager fluentd
-! systemctl status --wait --no-pager td-agent
+systemctl status --no-pager fluentd
+! systemctl status --no-pager td-agent
 
 # Test: restoring td-agent service alias
 sudo systemctl unmask td-agent
 sudo systemctl enable --now fluentd
 
-systemctl status --wait --no-pager td-agent
-systemctl status --wait --no-pager fluentd
+systemctl status --no-pager td-agent
+systemctl status --no-pager fluentd
 
 # Test: config migration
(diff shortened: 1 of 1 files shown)


daipom:
Note: `$ man systemctl` on Ubuntu 22.04.

```
       --no-block
           Do not synchronously wait for the requested operation to finish. If this is not specified, the job will be verified, enqueued and systemctl will wait until the unit's start-up is completed. By passing this argument, it is
           only verified and enqueued. This option may not be combined with --wait.
```

```
       --wait
           Synchronously wait for started units to terminate again. This option may not be combined with --no-block. Note that this will wait forever if any given unit never terminates (by itself or by getting stopped explicitly);
           particularly services which use "RemainAfterExit=yes".

           When used with is-system-running, wait until the boot process is completed before returning.
```

ashie (approved):
LGTM

I couldn't understand the document what `--wait` does, so I read the source code:
```console
aho@Ryzen:~/src/systemd-249.11/src/systemctl$ grep -rn arg_wait
systemctl.c:70:bool arg_wait = false;
systemctl.c:634:                        arg_wait = true;
systemctl.c:937:        if (arg_wait && arg_no_block)
systemctl.c:949:                if (arg_wait)
systemctl-start-unit.c:274:        if (arg_wait && !STR_IN_SET(argv[0], "start", "restart"))
systemctl-start-unit.c:280:        r = acquire_bus(arg_wait ? BUS_FULL : BUS_MANAGER, &bus);
systemctl-start-unit.c:354:        if (arg_wait) {
systemctl-start-unit.c:396:        if (arg_wait) {
systemctl.h:54:extern bool arg_wait;
systemctl-is-system-running.c:44:        if (arg_wait) {
systemctl-is-system-running.c:57:                        arg_wait = false;
systemctl-is-system-running.c:70:        if (arg_wait && STR_IN_SET(state, "initializing", "starting")) {
```
It's only affected on `systemctl start` or `systemctl is-system-running`, not affected to `status`.

ashie (changes requested):
> By default, systemctl status waits until the unit's start-up is completed (without --no-block option).
--wait is an option about terminating, and has nothing to do with the start-up.

Could you revise this comment?
The option `--wait` is for blocking `systemctl start foobar` until the service `foobar` is stopped.

daipom:
Thanks for your review! I see!

The commit message and the issue comment is updated!

ashie (approved):
LGTM

daipom:
Thanks!

PR #527, opened 2023-07-18
msi: apply new package name to default config

daipom:


daipom:
There is no longer any reason to leave these paths as before, right?

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #529, opened 2023-07-18
Remove td-agent-ruby.conf

daipom:
It was used for prelink.
We don't use it since 66116d55088b4c055579d2ea3b326b48f0cc4d24.


ashie (approved):

daipom:
Thanks for your review!

PR #530, opened 2023-07-19
deb: tweak tests

daipom:
* Fix inverted tests using `!`, which has always succeeded.
* Simplifies.
* Add some test cases.

### DIFF
--- a/fluent-package/apt/systemd-test/update-from-v4.sh
+++ b/fluent-package/apt/systemd-test/update-from-v4.sh
@@ -25,7 +25,7 @@ sudo apt install -V -y \
 
 # Test: service status
 systemctl status --no-pager fluentd
-! systemctl status --no-pager td-agent
+(! systemctl status --no-pager td-agent)
 
 # Test: restoring td-agent service alias
 sudo systemctl unmask td-agent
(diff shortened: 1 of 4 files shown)


fluent-package/apt/systemd-test/update-from-v4.sh
```diff
@@ -45,11 +45,17 @@ test -e /var/log/td-agent/td-agent.log
...
 test $(eval $env_vars && echo $USER) = "_fluentd"
 test $(eval $env_vars && echo $FLUENT_CONF) = "/etc/fluent/td-agent.conf"
-test $(eval $env_vars && echo $FLUENT_PLUGIN) = "/etc/fluent/plugin"
 test $(eval $env_vars && echo $FLUENT_PACKAGE_LOG_FILE) = "/var/log/fluent/td-agent.log"
+test $(eval $env_vars && echo $FLUENT_PLUGIN) = "/etc/fluent/plugin"
+test $(eval $env_vars && echo $FLUENT_SOCKET) = "/var/run/fluent/fluentd.sock"
+
+# Test: logs
+sleep 3
+! grep -q -e '\[warn\]' -e '\[error\]' -e '\[fatal\]' /var/log/fluent/td-agent.log
```
daipom, line 58: I noticed this should fail because the Fluentd works with the old config, which should output a warning log...

Why is this not failing...

daipom: I'm testing longer sleep on my fork.

daipom: I will remove this test from `update-from-v4` test after I find the cause.

daipom: It seems that `!` dose not work as expected.
Perhaps I misunderstand some basic knowledge about `!` and `set -e`.

daipom: `test -e` does not check some commands combined with logical operators.

The document of `set -e` from `man bash`:

>  The shell does not exit if the command that
                      fails  is  part  of  the command list immediately following a while or until keyword, part of the test following the if or elif reserved words, part of any command executed in a && or || list except the command
                      following the final && or ||, any command in a pipeline but the last, or if the command's return value is being inverted with !.

I don't understand why some commands such as `test` or `systemctl status` can be inverted and checked by `test -e`...

daipom: I used `if` and `exit` instead of `!`.

daipom: > It seems strange that instead of FLUENT_PACKAGE_LOG_FILE, TD_AGENT_LOG_FILE may exist in env_vars. 🤔

It is OK because the new environment variables are used!
The value is replaced by the migration process!

https://github.com/fluent/fluent-package-builder/blob/670ce2d5b592be5d5507df43f48f93af09df4ada/fluent-package/templates/package-scripts/fluent-package/deb/postinst#L106-L109

daipom: > `set -e` does not check some commands combined with logical operators.
> 
> The document of `set -e` from `man bash`:
> 
> > The shell does not exit if the command that
> > fails  is  part  of  the command list immediately following a while or until keyword, part of the test following the if or elif reserved words, part of any command executed in a && or || list except the command
> > following the final && or ||, any command in a pipeline but the last, or if the command's return value is being inverted with !.
> 
> I don't understand why some commands such as `test` or `systemctl status` can be inverted and checked by `set -e`...

This document is completely correct.
It was my misunderstanding that `test` and `systemctl status` can be inverted and checked by `set -e`,
All commands that are inverted by `!` can not exit by `set -e`.

daipom:
Fixed inverted tests with `!`.

daipom:
Rebased to the latest master to solve conflicts.

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #534, opened 2023-07-20
Update uri gem to suppress CVE-2023-36617

ashie:
ref: https://www.ruby-lang.org/en/news/2023/06/29/redos-in-uri-CVE-2023-36617/


daipom (approved):
LGTM. Thanks!

PR #535, opened 2023-07-20
rpm: tweak tests

daipom:
* Fix inverted tests using `!`, which has always succeeded.
* Simplifies.
* Add some test cases.


kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #538, opened 2023-07-24
deb: rpm: add log test for migration

daipom:

PR #539, opened 2023-07-24
deb: prevent Fluentd from running doubly

daipom:
Without stopping before `unmask` and `enable`, sometimes Fluentd runs doubly.


daipom:
This resolves the following unstable tests:

* #538
* #526

ashie (approved):

daipom:
Thanks for your review!

PR #541, opened 2023-07-24
deb: add fluent-lts-apt-source package for LTS

kenhys:
We need to maintain 2 channels such as

* Standard version
* LTS version

Each channel must have separate apt-line.

* https://packages.treasuredata.com/5/
* https://packages.treasuredata.com/lts/5

So, for LTS version, provide fluent-lts-apt-source.


daipom:
We will change here for each branch in the future.
So we don't touch here for now, right?

https://github.com/fluent/fluent-package-builder/blob/a5e12ab9b0846f725923227c5a1babd159b2028a/Rakefile#L22-L24

fluent-lts-apt-source/Rakefile
```diff
@@ -0,0 +1,146 @@
...
+require "time"
+
+# Debian
+CLEAN.include("apt/tmp")
+CLEAN.include("apt/build.sh")
+CLEAN.include("apt/env.sh")
+CLEAN.include("debian/tmp")
+CLOBBER.include("apt/repositories")
+
+class FluentdAptSourcePackageTask < PackageTask
```
daipom, line 14: For the same reason, we don't need to separate the class name, right?

kenhys: Thanks, I've fixed it.

kenhys:
> We will change here for each branch in the future.
So we don't touch here for now, right?

Yes, it will be fixed in LTS branch.

fluent-lts-apt-source/Rakefile
daipom: Oh! We needed to fix this! OK!

daipom (approved):
LGTM. Thanks!

PR #542, opened 2023-07-24
msi: keep TD_AGENT_TOPDIR

daipom:


kenhys (approved):
LGTM.


yum case failure has nothing to do whth  it.

daipom:
Thanks for your review!

PR #546, opened 2023-07-26
deb:rpm: stop editing settings for migration

daipom:
This fix stops editing `/etc/default/fluentd` or `/etc/sysconfig/fluentd` for migration.
We shouldn't touch the settings for migration.

For migration of the config file, we make a symlink:

* `/etc/fluent/fluent.conf` => `/etc/fluent/td-agent.conf`

We stop migrating the log filename.
If need to output logs to the old filename `td-agent.log`, user needs to set the environmental variable.


daipom:
> In this case, users should migrate the user service file manually. Since we provide symlinks `/etc/td-agent/` and `/var/log/td-agent/`, updating some paths should work:
> 
>     * `LD_PRELOAD`
> 
>     * `GEM_HOME`
> 
>     * `GEM_PATH`
> 
>     * `/opt/td-agent/bin/fluentd`

I assume the following script, for example.

```sh
sudo sed -i /etc/systemd/system/td-agent.service -e 's /opt/td-agent/lib/libjemalloc.so /opt/fluent/lib/libjemalloc.so g'
sudo sed -i /etc/systemd/system/td-agent.service -e 's /opt/td-agent/lib/ruby/gems/2.7.0/ /opt/fluent/lib/ruby/gems/3.2.0/ g'
sudo sed -i /etc/systemd/system/td-agent.service -e 's /opt/td-agent/lib/ruby/gems/2.7.0/ /opt/fluent/lib/ruby/gems/3.2.0/ g'
sudo sed -i /etc/systemd/system/td-agent.service -e 's /opt/td-agent/bin/fluentd /opt/fluent/bin/fluentd g'
echo "Alias=fluentd.service" | sudo tee -a /etc/systemd/system/td-agent.service
```

This should be done by users manually, I think.

daipom:
Recently, `ServerSpec` test often fails and we can't confirm the results of systemd tests.
To check the results, I ran the tests without `ServerSpec` test on my fork.
The tests for `yum` is stable.

* https://github.com/daipom/fluent-package-builder/actions/runs/5664397584

fluent-package/apt/systemd-test/update-from-v4-with-old-service-file.sh
```diff
@@ -0,0 +1,108 @@
...
+
+systemctl status --no-pager td-agent
+
+# Set custom service as "td-agent-server"
+sudo systemctl stop td-agent
+sudo cp /lib/systemd/system/td-agent.service /etc/systemd/system/td-agent-server.service
+sudo cp /etc/logrotate.d/td-agent /etc/logrotate.d/td-agent-server
+sudo cp /etc/td-agent/td-agent.conf /etc/td-agent/td-agent-server.conf
+sudo sed -i /etc/systemd/system/td-agent-server.service -e 's /etc/td-agent/td-agent.conf /etc/td-agent/td-agent-server.conf g'
+sudo sed -i /etc/systemd/system/td-agent-server.service -e 's /var/run/td-agent/td-agent.sock /var/run/td-agent-server/td-agent-server.sock g'
```
kenhys, line 28:
```suggestion
sudo sed -i /etc/systemd/system/td-agent-server.service -e 's,/var/run/td-agent/td-agent.sock,/var/run/td-agent-server/td-agent-server.sock,g'
```

Could you explicitly set non-space separator?
It must be exactly 1 space, but it may be bit hard to distinguish. ` `  `  `

daipom: I see! Thanks for the advice!
I will fix it.

daipom: Fixed.

kenhys:
By the way, how about testing with  `/etc/systemd/system/td-agent.d/override.conf`?

fluent-package/templates/package-scripts/fluent-package/deb/postinst
```diff
@@ -66,17 +66,23 @@ migration_from_v4_main_process() {
...
-            echo "Refer previous configuration <%= compat_service_name %>.conf ..."
-            sed -i"" /etc/default/<%= service_name %> -e "/FLUENT_CONF/c FLUENT_CONF=/etc/<%= package_dir %>/<%= compat_service_name %>.conf"
-            for d in $(ls /etc/<%= compat_package_dir %>); do
-                if [ ! "$d" = "plugin" -a ! "$d" = "<%= compat_service_name %>.conf" ]; then
-                    # except managed under deb files must be migrated
-                    mv -f /etc/<%= compat_package_dir %>/$d /etc/<%= package_dir %>/
-                fi
-            done
+        fi
+        for d in $(ls /etc/<%= compat_package_dir %>); do
```
kenhys, line 77: With this for ... do, it seems that above `if [ -f /etc/<%= compat_package_dir %>/%<= compat_servie_name %>.conf ]; then ` block may be redundant.

daipom:
> By the way, how about testing with `/etc/systemd/system/td-agent.d/override.conf`?

Thanks. I will separate the fixes about user service file tests from this PR.
I will consider it later!

daipom:
yum tests are all green in my fork(without serverspec test): https://github.com/daipom/fluent-package-builder/actions/runs/5666906256 

The Jammy test fails because some gem paths of v450 are different from other platform.
It is just a problem of tests.
I will separate these new tests from this PR since it is not directly related to this PR.

fluent-package/templates/package-scripts/fluent-package/deb/postinst
kenhys: I'm overlooked the previous block condition, sorry.

daipom:
> I will separate the fixes about user service file tests from this PR.

Done.

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #548, opened 2023-07-26
rpm: ensure whether old key has been imported

kenhys:
If you didn't import public key yet, verifying
with rpm -K fails (digests SIGNATURES NOT OK)

It help step how to fix missing import key.


daipom (approved):
LGTM.

kenhys:
Instead of `A || B && C`,  it should be `A || (B && C) ` to work as expected.

PR #551, opened 2023-07-27
rpm: do not always remove compatible symlink during upgrading

kenhys:
Before:

  v4 => v5 => upgrading... it will remove cpmpatible symlinks => v5.x

After:

  v4 => v5 => upgrading... do not remove compatible symlinks => v5.x

### DIFF
--- a/fluent-package/yum/fluent-package.spec.in
+++ b/fluent-package/yum/fluent-package.spec.in
@@ -266,17 +266,22 @@ fi
 
 %postun
 %systemd_postun_with_restart @SERVICE_NAME@.service
-if [ -h /usr/sbin/@COMPAT_SERVICE_NAME@ ]; then
-  rm -f /usr/sbin/@COMPAT_SERVICE_NAME@
-fi
-if [ -h /usr/sbin/@COMPAT_SERVICE_NAME@-gem ]; then
-  rm -f /usr/sbin/@COMPAT_SERVICE_NAME@-gem
-fi
-if [ -h /etc/@COMPAT_PACKAGE_DIR@ ]; then
-  rm -f /etc/@COMPAT_PACKAGE_DIR@
-fi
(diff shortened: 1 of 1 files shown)


kenhys:
Try upgrade from v4 => v5 => v5.

```
$ sudo yum install -y ./rpmbuild/RPMS/x86_64/fluent-package-5.0.0-2.el7.x86_64.rpm 
Failed to set locale, defaulting to C
Loaded plugins: fastestmirror
Examining ./rpmbuild/RPMS/x86_64/fluent-package-5.0.0-2.el7.x86_64.rpm: fluent-package-5.0.0-2.el7.x86_64
Marking ./rpmbuild/RPMS/x86_64/fluent-package-5.0.0-2.el7.x86_64.rpm as an update to fluent-package-5.0.0-1.el7.x86_64
Resolving Dependencies
--> Running transaction check
---> Package fluent-package.x86_64 0:5.0.0-1.el7 will be updated
---> Package fluent-package.x86_64 0:5.0.0-2.el7 will be an update
--> Finished Dependency Resolution

```

After upgrading, compatible symlink was kept.

```
$ ls -la /usr/sbin/td-agent*
lrwxrwxrwx. 1 root root 17 Jul 27 04:56 /usr/sbin/td-agent -> /usr/sbin/fluentd
lrwxrwxrwx. 1 root root 20 Jul 27 04:56 /usr/sbin/td-agent-gem -> /usr/sbin/fluent-gem
```

After removing fluent-package, symlink was removed (expected).

```
$ sudo yum remove fluent-package.x86_64 
Failed to set locale, defaulting to C
$ ls -la /usr/sbin/td*
ls: cannot access /usr/sbin/td*: No such file or directory
```

daipom (approved):
LGTM. Thanks!

PR #553, opened 2023-07-27
deb: drop compatible symlink when no upgrade mode

kenhys:
Before:

v4 => v5 => upgrading... it will remove cpmpatible symlinks => v5.x

After:

v4 => v5 => upgrading... do not remove compatible symlinks => v5.x

### DIFF
--- a/fluent-package/templates/package-scripts/fluent-package/deb/postrm
+++ b/fluent-package/templates/package-scripts/fluent-package/deb/postrm
@@ -38,11 +38,13 @@ if [ "$1" = "purge" ]; then
 	    userdel --remove --force <%= compat_service_name %>
 	fi
 fi
-if [ -h /usr/sbin/<%= compat_service_name %> ]; then
-    rm -f /usr/sbin/<%= compat_service_name %>
-fi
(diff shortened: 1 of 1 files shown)


kenhys:
```
$ sudo apt install -y ./fluent-package_5.0.0-2_amd64.deb 
Reading package lists... Done
```

After upgrading, symlink was kept.

```
$ ls -la /usr/sbin/td-agent*
lrwxrwxrwx 1 root root 17 Jul 27 05:29 /usr/sbin/td-agent -> /usr/sbin/fluentd
lrwxrwxrwx 1 root root 20 Jul 27 05:29 /usr/sbin/td-agent-gem -> /usr/sbin/fluent-gem
```

```
$ sudo apt remove fluent-package 
Reading package lists... Done
```

After removing fluent-package, symlink was removed.

```
$ ls -la /usr/sbin/td-agent*
ls: cannot access '/usr/sbin/td-agent*': No such file or directory
```

daipom (approved):
Thanks!

Note: We should fix the condition of removing config and log links as well in the future.

PR #556, opened 2023-07-28
apt: fix a typo about filename

kenhys:


kenhys:
It should not be overlooked. orz.

daipom (approved):
LGTM. I overlooked it too. Sorry.

PR #557, opened 2023-07-28
rpm: use sudo to import correctly

kenhys:


daipom (approved):
LGTM.

daipom:
Yum tests without the serverspec test: https://github.com/daipom/fluent-package-builder/actions/runs/5689353698

PR #558, opened 2023-07-31
rpm: update kafka related test case

kenhys:
Before:

* confluent 6.x series are already reaches EOL.
  See https://docs.confluent.io/platform/current/installation/versions-interoperability.html

After:

* Switch to confluent 7.4, it will be supported until May 3, 2025.
* Use newer jdk for testing to avoid issue which is derived exception from
  obsolete one.
* Use newer command line option for kafka-topics. (--zookeeper option
is deprecated)

### DIFF
--- a/fluent-package/yum/serverspec-test.sh
+++ b/fluent-package/yum/serverspec-test.sh
@@ -113,21 +113,23 @@ EOF
 		exit 1
 	    fi
 	done
+	# Allow connection to kafka server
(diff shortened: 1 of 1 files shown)


kenhys:
waiting CI...

kenhys:
Fixed unbound variables.

kenhys:
rockylinux test case has passed.

image

daipom (approved):
Wow! Thanks for this fix!
LGTM.

PR #561, opened 2023-08-02
CHANGELOG.md: Add notice about v5.0.0

ashie:
We decided to treat v5.0.0 RC version, not GA.
So we should notice it to uers here.


kenhys (approved):
LGTM.

PR #569, opened 2023-08-15
CHANGELOG: fluent-plugin-windows-eventlog is not updated for v5.0.0

daipom:


daipom:
`fluent-plugin-windows-eventlog` is already v0.8.3 in td-agent v4.5.0.

https://github.com/fluent/fluent-package-builder/blob/v4.5.0/td-agent/Gemfile#L84
https://github.com/fluent/fluent-package-builder/blob/v4.5.0/td-agent/Gemfile.lock#L162
https://github.com/fluent/fluent-package-builder/blob/v4.5.0/td-agent/Gemfile.lock#L302

daipom:
This diff shows confusing results.
It may be because the path was changed.
We need to be careful with this :cry: 

https://github.com/fluent/fluent-package-builder/compare/v4.5.0...v5.0.0#diff-505194dbf36462a7c6dd7ff691a995d2ef382ed826c4b33877def49051df1550R88

kenhys (approved):
Surely PR#522 doesn't update it.

I've overlooked it.

Thanks.

daipom:
Thanks for your review!

PR #570, opened 2023-08-15
deb: refactor postrm script

kenhys:
Before:
 * Missing removing /var/run
 * /var/run/td-agent is kept when upgraded from v4.

After:
 * group procedure to function
 * purge /var/run during removing package

### DIFF
--- a/fluent-package/templates/package-scripts/fluent-package/deb/postrm
+++ b/fluent-package/templates/package-scripts/fluent-package/deb/postrm
@@ -4,7 +4,7 @@ set -e
 
 . /usr/share/debconf/confmodule
 
-if [ "$1" = "purge" ]; then
+purge_conf_files() {
 	rm -f /etc/default/<%= compat_service_name %>
 	rm -f /etc/default/<%= service_name %>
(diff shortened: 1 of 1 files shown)


fluent-package/templates/package-scripts/fluent-package/deb/postrm
```diff
@@ -30,21 +40,41 @@ if [ "$1" = "purge" ]; then
...
 	    userdel --remove --force _<%= service_name %>
 	fi
 	if getent passwd <%= compat_service_name %>; then
 	    userdel --remove --force <%= compat_service_name %>
 	fi
-fi
-if [ ! "$1" = "upgrade" ]; then
+}
+
+purge_symlinks() {
```
daipom, line 54: This name seems a little confusing since there are other symlinks for the config and the logs.
More specific name such as `purge_bin_symlinks` would be better I think.

daipom (approved):
Basically, this looks good to me.
I have commented on one point about the name of a function.

fluent-package/templates/package-scripts/fluent-package/deb/postrm
kenhys: Thanks, I've fixed it.

PR #575, opened 2023-08-28
td-agent v4.5.1

ashie:


ashie:
Built packages are already uploaded.

ashie:
gems are aligned to 5.0.1, so they aren't current latest.

kenhys (approved):
LGTM.

PR #576, opened 2023-08-29
deb: fix upstream and path to licenses

kenhys:
* deb: fix Upstream-Name: field
* deb: fix path of licenses to /opt/fluent/LICENSES/

PR #577, opened 2023-08-29
deb: set correct contact

kenhys:
Upstream accepts URIs.

See https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#upstream-contact-field

### DIFF
--- a/fluent-apt-source/debian/copyright
+++ b/fluent-apt-source/debian/copyright
@@ -1,6 +1,6 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: Fluentd Project
-Upstream-Contact: Fluentd developers <fluentd@googlegroups.com>
+Upstream-Contact: https://github.com/fluent/fluentd
 
 Files: *
 Copyright: 2020 Kentaro Hayashi <hayashi@clear-code.com>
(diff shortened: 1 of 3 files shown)


kenhys:
Follow up of #576

daipom (approved):
Thanks!
This fix looks good to me, but I noticed `Fluentd developers <fluentd@googlegroups.com>` is used too for `Mainainer` in the `control` files.
However, I can't find a document that says we can use a URL for it.

https://wiki.debian.org/Packaging/Intro#Step_3:_Add_the_Debian_packaging_files

> Maintainer
>    The name and e-mail address of the person responsible for the package. 

Should we keep `Fluentd developers <fluentd@googlegroups.com>` for `Maintainer` as it is because we have no choice?

https://github.com/fluent/fluent-package-builder/blob/fd741de75b7db48b055bb4b45f5fb4efd30a106c/fluent-package/debian/control#L4

https://github.com/fluent/fluent-package-builder/blob/fd741de75b7db48b055bb4b45f5fb4efd30a106c/fluent-lts-apt-source/debian/control#L4

https://github.com/fluent/fluent-package-builder/blob/fd741de75b7db48b055bb4b45f5fb4efd30a106c/fluent-apt-source/debian/control#L4

kenhys:
Currently, URI is not accepted.

https://www.debian.org/doc/debian-policy/ch-controlfields.html#maintainer

Issue #579, opened 2023-08-30
Switch package signing key


kenhys:
# Problem

Currently, we can't switch signing key yet.

# Expected

Switch package signing key after releasing fluent-package v5 series.

Fluentd developers (Fluent Package Official Signing Key)
B40948B6A3B80E90F40E841F977D7A0943FA320E

# Background

When releasing fluent-package v5.0.0, v5.0.1, we still use with old signing key because existing td-agent user can upgrade by 
even though manually editing apt-line fluentd.sources.

https://github.com/fluent/fluent-package-builder/blob/master/fluent-package/manage-fluent-repositories.sh#L36

* If user install package via install script, new signing key is also installed. 
  * deb: /usr/share/keyrings/fluent-archive-keyring.gpg
  * rpm: import via gpg --import (old and new key)

So switching signing key affects only existing td-agent users who manually point to fluent-package apt line.

PR #580, opened 2023-08-30
deb: switch package signing key

kenhys:
Switch signing key when fluent-apt-source package
was installed to most of machines.

As we can't measure such a metrics, just switch after a while. 
(It is important to determine "WHEN" we should switch to.

Before:

  Treasure Data, Inc (Treasure Agent Official Signing key)

After:

  Fluentd developers (Fluent Package Official Signing Key)

Closes: #579


daipom (approved):
LGTM! Thanks!

PR #583, opened 2023-09-01
ci: use latest td-agent package

kenhys:


daipom:
Thanks!
I think we need to fix `commonvar` file for yum as well as apt.

https://github.com/fluent/fluent-package-builder/blob/master/fluent-package/yum/systemd-test/commonvar.sh

kenhys:
Thanks, fixed it.

daipom (approved):
LTGM. Waiting CI...

PR #586, opened 2023-09-11
github: add missing bookworm as target

kenhys:
* github: drop redundant --mirror option

    Use pre-configured chroot apt-line, so --mirror is redundant.
    Without it, it causes the following error on bookworm:

      The following packages have unmet dependencies:
       fluent-package : Depends: libncurses6 (>= 6) but it is not installable
      W: Target Packages (contrib/binary-amd64/Packages) is configured multiple times in /etc/apt/sources.list:2 and /etc/apt/sources.list.d
      W: Target Packages (contrib/binary-all/Packages) is configured multiple times in /etc/apt/sources.list:2 and /etc/apt/sources.list.d/t
      W: Target Translations (contrib/i18n/Translation-en) is configured multiple times in /etc/apt/sources.list:2 and /etc/apt/sources.list

 github: add more .service check after removing

    After removing fluent-package, the status of service files seems
    different.

    bookworm:
      no /etc/systemd/system/fluentd.service
      dead symlink /etc/systemd/system/multi-user.target.wants/luentd.service
      dead /etc/systemd/system/td-agent.service

    bullseye:
      dead /etc/systemd/system/fluentd.service
      dead symlink /etc/systemd/system/multi-user.target.wants/luentd.service
      dead /etc/systemd/system/td-agent.service

### DIFF
--- a/fluent-package/apt/systemd-test/update-from-v4.sh
+++ b/fluent-package/apt/systemd-test/update-from-v4.sh
@@ -72,4 +72,8 @@ sudo apt remove -y fluent-package
 (! systemctl status --no-pager fluentd)
 
 test -h /etc/systemd/system/td-agent.service
+(! test -s /etc/systemd/system/td-agent.service)
 test -h /etc/systemd/system/fluentd.service
+(! test -s /etc/systemd/system/fluentd.service)
+test -h /etc/systemd/system/multi-user.target.wants/fluentd.service
+(! test -s /etc/systemd/system/multi-user.target.wants/fluentd.service)
--- a/.github/workflows/apt-arm.yml
+++ b/.github/workflows/apt-arm.yml
@@ -9,12 +9,16 @@ jobs:
       matrix:
         label:
           - Debian GNU/Linux bullseye arm64
+          - Debian GNU/Linux bookworm arm64
           - Ubuntu Focal arm64
           - Ubuntu Jammy arm64
         include:
           - label: Debian GNU/Linux bullseye arm64
             rake-job: debian-bullseye
             test-docker-image: arm64v8/debian:bullseye
+          - label: Debian GNU/Linux bookworm arm64
+            rake-job: debian-bookworm
+            test-docker-image: arm64v8/debian:bookworm
           - label: Ubuntu Focal arm64
             rake-job: ubuntu-focal
             rake-options: LINTIAN=no
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -13,12 +13,16 @@ jobs:
       matrix:
         label:
           - Debian GNU/Linux bullseye amd64
+          - Debian GNU/Linux bookworm amd64
           - Ubuntu Focal amd64
           - Ubuntu Jammy amd64
         include:
           - label: Debian GNU/Linux bullseye amd64
             rake-job: debian-bullseye
             test-docker-image: debian:bullseye
+          - label: Debian GNU/Linux bookworm amd64
+            rake-job: debian-bookworm
+            test-docker-image: debian:bookworm
           - label: Ubuntu Focal amd64
             rake-job: ubuntu-focal
             test-docker-image: ubuntu:focal
@@ -97,6 +101,7 @@ jobs:
       matrix:
         distribution:
           - debian-bullseye
+          - debian-bookworm
           - ubuntu-focal
           - ubuntu-jammy
     steps:
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,6 +11,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       :box => "bento/debian-11",
     },
     {
+      :id => "debian-bookworm",
+      :box => "bento/debian-12",
+    },
+    {
       :id => "ubuntu-focal",
       :box => "bento/ubuntu-20.04",
     },
--- a/fluent-package/apt/install-test.sh
+++ b/fluent-package/apt/install-test.sh
@@ -37,6 +37,10 @@ for conf_path in /etc/td-agent/td-agent.conf /etc/fluent/fluentd.conf; do
     fi
 done
 
+if [ "${code_name}" == "bookworm" ]; then
+    echo "As bookworm is not published for v4, so package upgrade install check for ${code_name} is disabled"
+    exit 0
+fi
 # TODO: Remove it when v5 repository was deployed
 apt install -y curl
 curl -O https://packages.treasuredata.com/4/${distribution}/${code_name}/pool/contrib/f/fluentd-apt-source/fluentd-apt-source_2020.8.25-1_all.deb
--- a/fluent-package/apt/piuparts-test.sh
+++ b/fluent-package/apt/piuparts-test.sh
@@ -25,6 +25,5 @@ echo "deb [signed-by=/usr/share/keyrings/td-agent-archive-keyring.gpg] https://p
 rm -rf $CHROOT/opt
 piuparts --distribution=${code_name} \
 	 --existing-chroot=${CHROOT} \
-	 --mirror="https://packages.treasuredata.com/5/${distribution}/${code_name}/ ${code_name} contrib" \
 	 --skip-logrotatefiles-test \
 	 /tmp/*_${architecture}.deb
--- a/fluent-package/apt/piuparts-test.sh
+++ b/fluent-package/apt/piuparts-test.sh
@@ -21,10 +21,10 @@ chmod 644 $CHROOT/usr/share/keyrings/td-agent-archive-keyring.gpg
 chroot $CHROOT apt install -V -y libyaml-0-2
 package=${repositories_dir}/${distribution}/pool/${code_name}/${channel}/*/*/*_${architecture}.deb
 cp ${package} /tmp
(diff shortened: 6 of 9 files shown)


daipom (changes requested):
Thanks!
I think we need to add https://app.vagrantup.com/bento/boxes/debian-12 to the Vagrant file.

https://github.com/fluent/fluent-package-builder/blob/2b9fd5eef0655b70dfc7a6f0a6cc506bea44dad0/Vagrantfile#L8-L13

```ruby
    {
      :id => "debian-bookworm",
      :box => "bento/debian-12",
    },
```

kenhys:
Oops, I've forgot to add it. Thanks.

daipom (approved):
Thanks!
It looks good to me if there is no problem with the CI result.

daipom:
https://github.com/fluent/fluent-package-builder/actions/runs/6144619825/job/16670375606?pr=586

Hmm, something seems wrong.

kenhys:
I'll check it.

kenhys:
It is because bookworm was supported since v5. (no bookworm deb for v4)

daipom:
Thanks!
In addition, I think this code does not work as expected.

https://github.com/fluent/fluent-package-builder/blob/cb795a5253e18f05a754ca9d41cef4aa2dbf8536/fluent-package/apt/pkgsize-test.sh#L14-L22

`$code_name` is from this.

https://github.com/fluent/fluent-package-builder/blob/cb795a5253e18f05a754ca9d41cef4aa2dbf8536/fluent-package/apt/commonvar.sh#L1

It seems that we should use `$CODE_NAME` instead of `$code_name`.

It is so confusing, so it may be better to change the name `$CODE_NAME`...

daipom:
> Thanks! In addition, I think this code does not work as expected.
> 
> https://github.com/fluent/fluent-package-builder/blob/cb795a5253e18f05a754ca9d41cef4aa2dbf8536/fluent-package/apt/pkgsize-test.sh#L14-L22
> 
> `$code_name` is from this.
> 
> https://github.com/fluent/fluent-package-builder/blob/cb795a5253e18f05a754ca9d41cef4aa2dbf8536/fluent-package/apt/commonvar.sh#L1
> 
> It seems that we should use `$CODE_NAME` instead of `$code_name`.
> 
> It is so confusing, so it may be better to change the name `$CODE_NAME`...

Hmm, what I'm saying seems partially wrong...

It seems not wrong to use `$code_name` here...

I don't understand why there is two variarables and this condition does not work though...

kenhys:
checking why...

```
 0m5.3s DEBUG: Starting command: ['chroot', '/tmp/tmpr3kvd8i9', 'apt-get', '-y', '--allow-downgrades', 'install', './tmp/fluent-package-dbgsym_5.0.1-1_amd64.deb', './tmp/fluent-package_5.0.1-1_amd64.deb']
0m5.4s DUMP: 
  Reading package lists...
  Building dependency tree...
  Reading state information...
  Some packages could not be installed. This may mean that you have
  requested an impossible situation or if you are using the unstable
  distribution that some required packages have not yet been created
  or been moved out of Incoming.
  The following information may help to resolve the situation:
  
  The following packages have unmet dependencies:
   fluent-package : Depends: libncurses6 (>= 6) but it is not installable
  W: Target Packages (contrib/binary-amd64/Packages) is configured multiple times in /etc/apt/sources.list:2 and /etc/apt/sources.list.d/td.list:1
  W: Target Packages (contrib/binary-all/Packages) is configured multiple times in /etc/apt/sources.list:2 and /etc/apt/sources.list.d/td.list:1
  W: Target Translations (contrib/i18n/Translation-en) is configured multiple times in /etc/apt/sources.list:2 and /etc/apt/sources.list.d/td.list:1
  E: Unable to correct problems, you have held broken packages.
```

kenhys:
Actually, depends on libncurses6.

```
 apt depends fluent-package
fluent-package
  PreDepends: adduser
  Depends: libc6 (>= 2.35)
  Depends: libcrypt1 (>= 1:4.1.0)
  Depends: libffi8 (>= 3.4)
  Depends: libgcc-s1 (>= 4.2)
  Depends: libgmp10 (>= 2:6.2.1+dfsg1)
  Depends: libncurses6 (>= 6)
  Depends: libreadline8 (>= 6.0)
  Depends: libssl3 (>= 3.0.0)
  Depends: libtinfo6 (>= 6)
  Depends: libyaml-0-2
  Depends: zlib1g (>= 1:1.2.11.dfsg)
  Breaks: td-agent (<< 5.0.0-1)
  Replaces: td-agent (<< 5.0.0-1)
```

And installed ncurses6.

```
dpkg -l |grep libncurses
ii  libncurses6:amd64         6.4-4                          amd64        shared libraries for terminal handling
ii  libncursesw6:amd64        6.4-4                          amd64        shared libraries for terminal handling (wide character support)
```

kenhys:
checking current repos:

```
apt-cache madison fluent-package
fluent-package |    5.0.1-1 | https://packages.treasuredata.com/5/debian/bookworm bookworm/contrib amd64 Packages
fluent-package |    5.0.0-1 | https://packages.treasuredata.com/5/debian/bookworm bookworm/contrib amd64 Packages
```

kenhys:
`apt-get install -y --allow-downgrades ./fluent-package-dbgsym_5.0.1-1_amd64.deb ./fluent-package_5.0.1-1_amd64.deb ` itself is not problem on bookworm container.

kenhys:
Got it, `--mirror` is redundant in this case.

kenhys:
waiting CI..

kenhys:
Fixed bookworm failure. yay!

kenhys:
a bit confused behavior is observed.


bullseye:
before: no /etc/systemd/system/fluentd.service
after(apt remove): /etc/systemd/system/fluentd.service => /dev/null (dead link)

bookworm:
before: no /etc/systemd/system/fluentd.service
after(apt remove): no /etc/systemd/system/fluentd.service, instead, /etc/systemd/system/multi-user.target.wants/fluentd.service=>/lib/systemd/system/fluentd.service (dead link)

daipom (commented):
Hmm, I tested installing v5 in bookworm, and I noticed the service link file becomes `/etc/systemd/system/td-agent.service`, not `/etc/systemd/system/fluentd.service`...

Sorry, this is an intended behavior. We need this for the service alias.

kenhys:
I've updated description and test case a bit.

daipom:
Hmm, certainly, there seems to be a specification change between debian-bullseye and debian-bookworm...

I think the dead link `/etc/systemd/system/fluentd.service` is to provide access to the service after removing.

We can still access the service after removing the package in bullseye:

```console
vagrant@debian-11:/etc/systemd/system$ systemctl status fluentd
● fluentd.service
     Loaded: masked (Reason: Unit fluentd.service is masked.)
     Active: inactive (dead) since Wed 2023-09-13 09:16:49 UTC; 13s ago
   Main PID: 11665 (code=exited, status=0/SUCCESS)
        CPU: 896ms
vagrant@debian-11:/etc/systemd/system$ systemctl status td-agent
● td-agent.service
     Loaded: masked (Reason: Unit td-agent.service is masked.)
     Active: inactive (dead)
```

We can't access the service anymore after removing the package in bookworm:

```console
vagrant@debian-12:~$ systemctl status fluentd
Unit fluentd.service could not be found.
vagrant@debian-12:~$ systemctl status td-agent
Unit td-agent.service could not be found.
```

I think this is not an issue of this package but of the specification of bookworm.

daipom:
> I've updated description and test case a bit.

Thanks, this change looks reasonable to me.
I also think it might not be necessary to test the remained service file link...

It seems that we should also fix `update-from-v4.sh`:

https://github.com/fluent/fluent-package-builder/blob/2b9fd5eef0655b70dfc7a6f0a6cc506bea44dad0/fluent-package/apt/systemd-test/update-from-v4.sh#L69-L75

fluent-package/apt/systemd-test/update-from-v4.sh
```diff
@@ -4,6 +4,11 @@ set -exu
 
 . $(dirname $0)/../commonvar.sh
 
+if [ "${code_name}" == "bookworm" ]; then
+    echo "As bookworm is not published for v4, so upgrade from v4 check is disabled"
+    exit 0
+fi
+
```
daipom, line 11: How about managing which tests should be run in `test.sh` together instead of each test?

```diff
diff --git a/fluent-package/apt/systemd-test/test.sh b/fluent-package/apt/systemd-test/test.sh
index fd862c6..e45005b 100755
--- a/fluent-package/apt/systemd-test/test.sh
+++ b/fluent-package/apt/systemd-test/test.sh
@@ -19,11 +19,16 @@ fi
 set -eu
 
 test_filenames=(
-    update-from-v4.sh
     update-to-next-version.sh
-    update-to-next-version-with-backward-compat-for-v4.sh
 )
 
+if [ $vm != "debian-bookworm" ]; then
+    test_filenames+=(
+        update-from-v4.sh
+        update-to-next-version-with-backward-compat-for-v4.sh
+    )
+fi
+
 for apt_repo_type in local v5 lts; do
     echo -e "\nRun test: $apt_repo_type\n"
     vagrant up $vm
```

kenhys: It may be reasonable to do so. Thanks.

kenhys:
waiting CI...

kenhys:
read_nonblock frequently occurs, :thinking:

kenhys:
> read_nonblock frequently occurs, 🤔

Try to search workaround in another issue/pr.

kenhys:
With some retry, all check has passed now.

daipom (approved):
LGTM! Thanks so much!

Issue #585, opened 2023-09-11
Improve downloading artifiacts procedure


kenhys:
# Problem

downloading artifacts procedure is not automated yet.

# Expected

automate downloading prebuilt artifacts and integrate repository easily.

kenhys:
If `head_branch` and `head_sha` is determined, we can filter response.

kenhys:
changed interface to more simple one such as:

```
$ ./manage-fluent-repositories.sh download-artifacts https://github.com/fluent/fluent-package-builder/pull/587
```

PR #587, opened 2023-09-11
package: add command to download artifacts from GitHub

kenhys:
Closes: #585

Specify the PR number like this. (Only the number or the full URL is valid too)

$ ./manage-fluent-repositories.sh download-artifacts pull/587

NOTE: You must set GITHUB_ACCESS_TOKEN in advance. See Fine-grained personal access tokens via
https://github.com/settings/tokens?type=beta

### DIFF
--- a/fluent-package/manage-fluent-repositories.sh
+++ b/fluent-package/manage-fluent-repositories.sh
@@ -50,6 +50,8 @@ case $COMMAND in
 	# Allow copying the URL from browser's URL bar.
 	PULL_REQUEST_URL=$2
 	PULL_NUMBER=${PULL_REQUEST_URL##*/}
+	read -rsp "Please enter your GitHub personal access token: " GITHUB_ACCESS_TOKEN
+	echo
 	;;
     *)
 	FLUENT_RELEASE_PROFILE=$2
--- a/fluent-package/manage-fluent-repositories.sh
+++ b/fluent-package/manage-fluent-repositories.sh
@@ -24,7 +24,7 @@ Example:
   $ $0 upload release-td-agent /tmp/td-agent-release
   $ $0 deb /tmp/td-agent-release 4.2.0
   $ $0 rpm /tmp/td-agent-release 4.2.0
-  $ $0 download-artifacts https://github.com/fluent/fluent-package-builder/pull/587
+  $ $0 download-artifacts pull/587
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


kenhys:
still in WIP, it may be a bit complicated.

fluent-package/manage-fluent-repositories.sh
```diff
@@ -24,6 +24,7 @@ Example:
   $ $0 upload release-td-agent /tmp/td-agent-release
   $ $0 deb /tmp/td-agent-release 4.2.0
   $ $0 rpm /tmp/td-agent-release 4.2.0
+  $ $0 download-artifacts https://github.com/fluent/fluent-package-builder/pull/587
```
ashie, line 27: The base URL isn't used in actual, so the example might be better to simplify like this:
`$0 download-artifacts 587` or `$0 download-artifacts pull/587`

kenhys: Thanks, It seems reasonable. I'll fix it.

daipom:
I will see this next week.

fluent-package/manage-fluent-repositories.sh
```diff
@@ -44,6 +45,12 @@ case $COMMAND in
 	    exit 1
 	fi
 	;;
+    download-artifacts)
+	# Given URL will not be used. Just use the number of pull request.
+	# Allow copying the URL from browser's URL bar.
+	PULL_REQUEST_URL=$2
+	PULL_NUMBER=${PULL_REQUEST_URL##*/}
```
daipom, line 52: How should we specify `GITHUB_ACCESS_TOKEN`?
That value is very secret, so we should be careful about specifying it.

I think this would be convenient.

```suggestion
	PULL_REQUEST_URL=$2
	PULL_NUMBER=${PULL_REQUEST_URL##*/}
	read -rsp "Please enter your GitHub personal access token: " GITHUB_ACCESS_TOKEN
	echo
```

What do you think?

kenhys: Thanks, it seems reasonable change.

daipom:
@kenhys DCO is missing!

kenhys:
Fixed it.

daipom (approved):
LGTM. Thanks!

daipom:
I have modified the description of PR a little bit.

PR #589, opened 2023-09-14
msi: support path which contains space or parenthesis

kenhys:
When installing fluent-package to c:\Program Files, the space
character must be escaped correctly.

* sources.wxs
  * For xml, double quote (") must be &quot;
  * to escape double quote in literal, you must escape with ^&quote;
    it is treated as ^", so it is properly quoted to execute:

  "cmd" /c "fsutil ... "c:\Program Files .." "c:\Program Files...""

* fluentd.bat,fluent-gem.bat
  * set FOO="..." assign extra double quote, it's troublesome because
  extra double quote will be included when concatenating path.
  The only way to avoid this problem is set "FOO=..." to assign it.
  you can assign to FOO without extra double quote.
  * to get RUBY_VERSION, use "usebackq" and execute command with
    escaped command literal explicitly.
     passed `^""C:\..." ....^"` literal are evaluated ^""C:\..."^".

### DIFF
--- a/fluent-package/msi/assets/fluent-gem.bat
+++ b/fluent-package/msi/assets/fluent-gem.bat
@@ -1,12 +1,12 @@
 @echo off
 if "%~nx0" == "td-agent-gem.bat" (
-  set FLUENT_PACKAGE_TOPDIR=%~dp0..\
+  set "FLUENT_PACKAGE_TOPDIR=%~dp0..\"
 ) else (
-  set FLUENT_PACKAGE_TOPDIR=%~dp0
+  set "FLUENT_PACKAGE_TOPDIR=%~dp0"
 )
-set FLUENT_PACKAGE_BINDIR=%FLUENT_PACKAGE_TOPDIR%bin
-set PATH="%FLUENT_PACKAGE_BINDIR%";%PATH%
-for /f "usebackq" %%i in (`%FLUENT_PACKAGE_BINDIR%\ruby.exe -rrbconfig -e "print RbConfig::CONFIG['ruby_version']"`) do set RUBY_VERSION=%%i
-set GEM_HOME=%FLUENT_PACKAGE_TOPDIR%lib\ruby\gems\%RUBY_VERSION%
-set GEM_PATH=%FLUENT_PACKAGE_TOPDIR%lib\ruby\gems\%RUBY_VERSION%
-%FLUENT_PACKAGE_BINDIR%\fluent-gem %*
+set "FLUENT_PACKAGE_BINDIR=%FLUENT_PACKAGE_TOPDIR%bin"
+set PATH=%FLUENT_PACKAGE_BINDIR%;%PATH%
+for /f "usebackq" %%i in (`^""%FLUENT_PACKAGE_BINDIR%\ruby.exe" -rrbconfig -e "print RbConfig::CONFIG['ruby_version']"^"`) do set RUBY_VERSION=%%i
+set "GEM_HOME=%FLUENT_PACKAGE_TOPDIR%lib\ruby\gems\%RUBY_VERSION%"
+set "GEM_PATH=%FLUENT_PACKAGE_TOPDIR%lib\ruby\gems\%RUBY_VERSION%"
+"%FLUENT_PACKAGE_BINDIR%\fluent-gem" %*
--- a/fluent-package/msi/assets/fluentd.bat
+++ b/fluent-package/msi/assets/fluentd.bat
@@ -1,8 +1,8 @@
 @echo off
 if "%~nx0" == "td-agent.bat" (
-  set FLUENT_PACKAGE_TOPDIR=%~dp0..\
+  set "FLUENT_PACKAGE_TOPDIR=%~dp0..\"
 ) else (
-  set FLUENT_PACKAGE_TOPDIR=%~dp0
+  set "FLUENT_PACKAGE_TOPDIR=%~dp0"
 )
 
 @rem Convert path separator from backslash to forwardslash
@@ -11,9 +11,9 @@ set FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!
 
 set PATH=%FLUENT_PACKAGE_TOPDIR%bin;%PATH%
 set PATH=%FLUENT_PACKAGE_TOPDIR%;%PATH%
-set FLUENT_CONF=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/fluentd.conf
-set FLUENT_PLUGIN=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/plugin
-set FLUENT_PACKAGE_VERSION=%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb
+set "FLUENT_CONF=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/fluentd.conf"
+set "FLUENT_PLUGIN=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/plugin"
+set "FLUENT_PACKAGE_VERSION=%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb"
 for %%p in (%*) do (
     if "%%p"=="--version" (
         ruby "%FLUENT_PACKAGE_VERSION%"
--- a/fluent-package/msi/source.wxs
+++ b/fluent-package/msi/source.wxs
@@ -185,7 +185,7 @@
     <Property Id="InstallFluentdWinSvc" Value=" "/>
     <CustomAction Id="SetInstallFluentdWinSvcCommand"
                   Property="InstallFluentdWinSvc"
-                  Value="&quot;[FLUENTPROJECTLOCATION]fluentd.bat&quot; --reg-winsvc i --reg-winsvc-delay-start --reg-winsvc-fluentdopt &quot;-c [FLUENTPROJECTLOCATION]etc\fluent\fluentd.conf -o [FLUENTPROJECTLOCATION]fluentd.log&quot;"/>
+                  Value="&quot;[FLUENTPROJECTLOCATION]fluentd.bat&quot; --reg-winsvc i --reg-winsvc-delay-start --reg-winsvc-fluentdopt &quot;-c '[FLUENTPROJECTLOCATION]etc\fluent\fluentd.conf' -o '[FLUENTPROJECTLOCATION]fluentd.log'&quot;"/>
     <CustomAction Id="InstallFluentdWinSvc"
                   BinaryKey="WixCA"
                   DllEntry="WixQuietExec64"
@@ -196,7 +196,7 @@
     <Property Id="CreateCompatTdAgentBat" Value=" "/>
     <CustomAction Id="SetCreateCompatTdAgentBat"
                   Property="CreateCompatTdAgentBat"
-                  Value="&quot;cmd&quot; /c &quot;fsutil hardlink create [FLUENTPROJECTLOCATION]bin\td-agent.bat [FLUENTPROJECTLOCATION]fluentd.bat&quot;"/>
+                  Value="&quot;cmd&quot; /c &quot;fsutil hardlink create ^&quot;[FLUENTPROJECTLOCATION]bin\td-agent.bat^&quot; ^&quot;[FLUENTPROJECTLOCATION]fluentd.bat^&quot;&quot;"/>
     <CustomAction Id="CreateCompatTdAgentBat"
                   BinaryKey="WixCA"
                   DllEntry="WixQuietExec64"
@@ -206,7 +206,7 @@
     <Property Id="CreateCompatTdAgentGemBat" Value=" "/>
     <CustomAction Id="SetCreateCompatTdAgentGemBat"
                   Property="CreateCompatTdAgentGemBat"
-                  Value="&quot;cmd&quot; /c &quot;fsutil hardlink create [FLUENTPROJECTLOCATION]bin\td-agent-gem.bat [FLUENTPROJECTLOCATION]fluent-gem.bat&quot;"/>
+                  Value="&quot;cmd&quot; /c &quot;fsutil hardlink create ^&quot;[FLUENTPROJECTLOCATION]bin\td-agent-gem.bat^&quot; ^&quot;[FLUENTPROJECTLOCATION]fluent-gem.bat^&quot;&quot;"/>
     <CustomAction Id="CreateCompatTdAgentGemBat"
                   BinaryKey="WixCA"
                   DllEntry="WixQuietExec64"
@@ -217,7 +217,7 @@
     <Property Id="DeleteCompatTdAgentBat" Value=" "/>
     <CustomAction Id="SetDeleteCompatTdAgentBat"
                   Property="DeleteCompatTdAgentBat"
-                  Value="&quot;cmd&quot; /c &quot;del [FLUENTPROJECTLOCATION]bin\td-agent.bat&quot;"/>
+                  Value="&quot;cmd&quot; /c &quot;del ^&quot;[FLUENTPROJECTLOCATION]bin\td-agent.bat^&quot;&quot;"/>
     <CustomAction Id="DeleteCompatTdAgentBat"
                   BinaryKey="WixCA"
                   DllEntry="WixQuietExec64"
@@ -227,7 +227,7 @@
     <Property Id="DeleteCompatTdAgentGemBat" Value=" "/>
     <CustomAction Id="SetDeleteCompatTdAgentGemBat"
                   Property="DeleteCompatTdAgentGemBat"
-                  Value="&quot;cmd&quot; /c &quot;del [FLUENTPROJECTLOCATION]bin\td-agent-gem.bat&quot;"/>
+                  Value="&quot;cmd&quot; /c &quot;del ^&quot;[FLUENTPROJECTLOCATION]bin\td-agent-gem.bat^&quot;&quot;"/>
     <CustomAction Id="DeleteCompatTdAgentGemBat"
                   BinaryKey="WixCA"
                   DllEntry="WixQuietExec64"


kenhys:
Installed to `c:\Program Files`:

fluent-escape-space-fluent

Then check fluentdopt:

fluent-escape-space-fluentdopt

fluent-package/msi/source.wxs
```diff
@@ -196,7 +196,7 @@
     <Property Id="CreateCompatTdAgentBat" Value=" "/>
     <CustomAction Id="SetCreateCompatTdAgentBat"
                   Property="CreateCompatTdAgentBat"
-                  Value="&quot;cmd&quot; /c &quot;fsutil hardlink create [FLUENTPROJECTLOCATION]bin\td-agent.bat [FLUENTPROJECTLOCATION]fluentd.bat&quot;"/>
+                  Value="&quot;cmd&quot; /c &quot;fsutil hardlink create ^&quot;[FLUENTPROJECTLOCATION]bin\td-agent.bat^&quot; ^&quot;[FLUENTPROJECTLOCATION]fluentd.bat^&quot;&quot;"/>
```
daipom, line 199: Thanks for this fix!

I have a question.
What is the difference between `'` and `^&quot;`?
I understand what `^&quot;` means, but can't we use `'` here?

daipom: Oh, sorry. I understand.
In Command Prompt, we can't use `'` for strings...

daipom: `fluentdopt` is a String to be passed to Ruby.exe, so we can use `'`.
On the other hand, for the options to be handled in Command Prompt itself, we have to use espaped `"`.

daipom:
I'm checking the package behavior...

daipom:
I failed to install the package to `C:\Dummy Program Files (FOO)\`.
I'm checking the cause.

daipom:
I checked the log.
 `InstallFluentdWinSvc` is failing.

```cmd
$  msiexec /i fluent-package-5.0.1-x64.msi /l* installer.log
```

```
Action 12:39:40: InstallFluentdWinSvc. 
WixQuietExec64:  \fluent\..\ の使い方が誤っています。
WixQuietExec64:  Error 0x800700ff: Command line returned an error.
WixQuietExec64:  Error 0x800700ff: QuietExec64 Failed
WixQuietExec64:  Error 0x800700ff: Failed in ExecCommon method
CustomAction InstallFluentdWinSvc returned actual error code 1603 (note this may not be 100% accurate if translation happened inside sandbox)
Action ended 12:39:40: InstallFinalize. Return value 3.
Action 12:39:40: Rollback. Rolling back action:
Rollback: InstallFluentdWinSvc
Rollback: Copying new files
Rollback: Creating folders
Rollback: Updating component registration
Action ended 12:40:10: INSTALL. Return value 3.
Action ended 12:40:10: ExecuteAction. Return value 3.
Action 12:40:10: FatalError. 
Action start 12:40:10: FatalError.
Action 12:40:10: FatalError. Dialog created
Action ended 12:40:12: FatalError. Return value 2.
Action ended 12:40:12: INSTALL. Return value 3.
```

kenhys:
Are there `Executing op:` log? `/l*v` option will enable such feature.

kenhys:
> Are there Executing op: log? /l*v option will enable such feature.

Reproduced. got same log.

kenhys:
assets/fluentd.bat cause it.

daipom:
I think this is completely another issue.
I will create another issue later.
We should handle this separately from this PR.

* This rollback occurs when the path includes `()`.
* This is caused by the `delayedexpansion` mechanism of BATCH.

I think this is the smallest script to reproduce:

```
@echo off
if "a" == "b" (
  echo %~dp0
)
```

daipom:
Continue to check the behavior when the installed path including spaces.

daipom:
bca8ebf967b25d6097271cac35d468fcc088f7fd

I see!
Certainly, `"` will solve this problem.

> I think this is completely another issue. I will create another issue later. We should handle this separately from this PR.
> 
>     * This rollback occurs when the path includes `()`.
> 
>     * This is caused by the `delayed environment variable` mechanism of BATCH.
> 
> 
> I think this is the smallest script to reproduce:
> 
> `/path/to/foo(bar)/test.bat`
> 
> ```
> @echo off
> if "a" == "b" (
>   echo %~dp0
> )
> ```
> 
> ```
> $ .\test.bat
> \ was unexpected at this time.
> ```

Thanks!

`/path/to/foo(bar)/test.bat`

```batch
@echo off
if "a" == "a" (
  echo "%~dp0"
)
```

```console
$ test.bat
"C:\path\to\foo(bar)\"
```

daipom:
> Continue to check the behavior when the installed path including spaces.

Installed path: `C:\Dummy Program Files\`

Something seems wrong with `fluent-gem`.
Looks like calling `fluent-gem` breaking the environment.

`Fluent Package Command Prompt` with administrative privileges:

```cmd
C:\Windows\system32>fluentd --version
fluent-package 5.0.1 fluentd 1.16.2 (d5685ada81ac89a35a79965f1e94bbe5952a5d3a)

C:\Windows\system32>fluent-gem --version
'C:\Dummy' は、内部コマンドまたは外部コマンド、
操作可能なプログラムまたはバッチ ファイルとして認識されていません。
'C:\Dummy' は、内部コマンドまたは外部コマンド、
操作可能なプログラムまたはバッチ ファイルとして認識されていません。

C:\Windows\system32>fluent-gem --version
C:/Dummy Program Files/fluent/lib/ruby/3.2.0/rubygems.rb:263:in `find_spec_for_exe': can't find gem fluentd (>= 0.a) with executable fluent-gem (Gem::GemNotFoundException)
        from C:/Dummy Program Files/fluent/lib/ruby/3.2.0/rubygems.rb:282:in `activate_bin_path'
        from C:/Dummy Program Files/fluent/bin/fluent-gem:32:in `<main>'

C:\Windows\system32>fluentd --version
C:/Dummy Program Files/fluent/lib/ruby/3.2.0/rubygems.rb:263:in `find_spec_for_exe': can't find gem fluentd (>= 0.a) with executable fluentd (Gem::GemNotFoundException)
        from C:/Dummy Program Files/fluent/lib/ruby/3.2.0/rubygems.rb:282:in `activate_bin_path'
        from C:/Dummy Program Files/fluent/bin/fluentd:32:in `<main>'
```

kenhys:
Yes fluent-gem also has same issue.

daipom:
Oh, is it the same issue?
I don't completely understand the mechanism...
OK, I will check the behavior again after the new package is built!

kenhys:
f1dd645 has still issues...

kenhys:
It should be `delims=`.

kenhys:
Now fluent-gem will work.

fluent-escape-space-fluent-gem

daipom:
Thanks. I'm checking the package.

Finally, I understood the mechanism of `\ was unexpected at this time.` error.

> `/path/to/foo(bar)/test.bat`
> 
> ```
> @echo off
> if "a" == "b" (
>   echo %~dp0
> )
> ```
> 
> ```
> $ .\test.bat
> \ was unexpected at this time.
> ```

`delayed environment variable` has nothing to do with this error.

Is is simply because the `()` of the path breaks the `if ()` block.

In the above case, `echo %~dp0` will be `echo /path/to/foo(bar)/`.
So, the script will be equivalent to the following.

```bat
@echo off
if "a" == "b" (
  echo /path/to/foo(bar)/
)
```

The error occurs simlply becaues this script is broken.
We can workaround this by `"`.

fluent-package/msi/assets/fluent-gem.bat
```diff
@@ -1,12 +1,12 @@
...
 )
-set FLUENT_PACKAGE_BINDIR=%FLUENT_PACKAGE_TOPDIR%bin
-set PATH="%FLUENT_PACKAGE_BINDIR%";%PATH%
-for /f "usebackq" %%i in (`%FLUENT_PACKAGE_BINDIR%\ruby.exe -rrbconfig -e "print RbConfig::CONFIG['ruby_version']"`) do set RUBY_VERSION=%%i
-set GEM_HOME=%FLUENT_PACKAGE_TOPDIR%lib\ruby\gems\%RUBY_VERSION%
-set GEM_PATH=%FLUENT_PACKAGE_TOPDIR%lib\ruby\gems\%RUBY_VERSION%
-%FLUENT_PACKAGE_BINDIR%\fluent-gem %*
+set "FLUENT_PACKAGE_BINDIR=%FLUENT_PACKAGE_TOPDIR%bin"
+set PATH=%FLUENT_PACKAGE_BINDIR%;%PATH%
+for /f "delims=" %%i in ('^""%FLUENT_PACKAGE_BINDIR%\ruby.exe" -rrbconfig -e "print RbConfig::CONFIG['ruby_version']"^"') do set RUBY_VERSION=%%i
```
daipom, line 9: Sorry for my late response.
It looks good to me except for this line.

`for` of `batch` is very difficult for me. I read https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/for.

My understanding is that using `usebackq` is not wrong, and we don't need to specify `delims` for the purpose of this PR.

What we need would be to surround the entire command with `^"`.

```suggestion
for /f "usebackq" %%i in (`^""%FLUENT_PACKAGE_BINDIR%\ruby.exe" -rrbconfig -e "print RbConfig::CONFIG['ruby_version']"^"`) do set RUBY_VERSION=%%i
```

If we don't use `usebackq`, then this fix would be enough for this PR.

```suggestion
for /f %%i in ('^""%FLUENT_PACKAGE_BINDIR%\ruby.exe" -rrbconfig -e "print RbConfig::CONFIG['ruby_version']"^"') do set RUBY_VERSION=%%i
```

Let me know if I missed something.

daipom:
Hmm, I thought there was no problem, but I have noticed that I can not run Fluentd in Fluent Package Command Prompt...
I'm checking the cause.

```cmd
$ fluentd
C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/config.rb:43:in `initialize': Invalid argument @ rb_sysopen - "C:/Dummy Program Files (FOO)/fluent//etc/fluent/fluentd.conf" (Errno::EINVAL)
        from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/config.rb:43:in `open'
        from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/config.rb:43:in `build'
        from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/supervisor.rb:682:in `setup_global_logger'
        from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/supervisor.rb:624:in `configure'
        from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/command/fluentd.rb:351:in `<top (required)>'
        from <internal:C:/Dummy Program Files (FOO)/fluent/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:C:/Dummy Program Files (FOO)/fluent/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/bin/fluentd:15:in `<top (required)>'
        from C:/Dummy Program Files (FOO)/fluent/bin/fluentd:32:in `load'
        from C:/Dummy Program Files (FOO)/fluent/bin/fluentd:32:in `<main>'
```

fluent-package/msi/assets/fluentd.bat
```diff
@@ -11,12 +11,12 @@ set FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!
 
 set PATH=%FLUENT_PACKAGE_TOPDIR%bin;%PATH%
 set PATH=%FLUENT_PACKAGE_TOPDIR%;%PATH%
-set FLUENT_CONF=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/fluentd.conf
-set FLUENT_PLUGIN=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/plugin
-set FLUENT_PACKAGE_VERSION=%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb
+set FLUENT_CONF="%FLUENT_PACKAGE_TOPDIR%/etc/fluent/fluentd.conf"
+set FLUENT_PLUGIN="%FLUENT_PACKAGE_TOPDIR%/etc/fluent/plugin"
+set FLUENT_PACKAGE_VERSION="%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb"
```
daipom, line 16:
```suggestion
set "FLUENT_CONF=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/fluentd.conf"
set "FLUENT_PLUGIN=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/plugin"
set "FLUENT_PACKAGE_VERSION=%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb"
```

fluent-package/msi/assets/fluentd.bat
```diff
@@ -11,12 +11,12 @@ set FLUENT_PACKAGE_TOPDIR=!FLUENT_PACKAGE_TOPDIR:\=/!
...
-set FLUENT_CONF=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/fluentd.conf
-set FLUENT_PLUGIN=%FLUENT_PACKAGE_TOPDIR%/etc/fluent/plugin
-set FLUENT_PACKAGE_VERSION=%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb
+set FLUENT_CONF="%FLUENT_PACKAGE_TOPDIR%/etc/fluent/fluentd.conf"
+set FLUENT_PLUGIN="%FLUENT_PACKAGE_TOPDIR%/etc/fluent/plugin"
+set FLUENT_PACKAGE_VERSION="%FLUENT_PACKAGE_TOPDIR%/bin/fluent-package-version.rb"
 for %%p in (%*) do (
     if "%%p"=="--version" (
-        ruby "%FLUENT_PACKAGE_VERSION%"
+        ruby %FLUENT_PACKAGE_VERSION%
```
daipom, line 19:
```suggestion
        ruby "%FLUENT_PACKAGE_VERSION%"
```

fluent-package/msi/assets/fluentd.bat
daipom: > Hmm, I thought there was no problem, but I have noticed that I can not run Fluentd in Fluent Package Command Prompt... I'm checking the cause.
> 
> ```batchfile
> $ fluentd
> C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/config.rb:43:in `initialize': Invalid argument @ rb_sysopen - "C:/Dummy Program Files (FOO)/fluent//etc/fluent/fluentd.conf" (Errno::EINVAL)
>         from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/config.rb:43:in `open'
>         from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/config.rb:43:in `build'
>         from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/supervisor.rb:682:in `setup_global_logger'
>         from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/supervisor.rb:624:in `configure'
>         from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/lib/fluent/command/fluentd.rb:351:in `<top (required)>'
>         from <internal:C:/Dummy Program Files (FOO)/fluent/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
>         from <internal:C:/Dummy Program Files (FOO)/fluent/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
>         from C:/Dummy Program Files (FOO)/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.16.2/bin/fluentd:15:in `<top (required)>'
>         from C:/Dummy Program Files (FOO)/fluent/bin/fluentd:32:in `load'
>         from C:/Dummy Program Files (FOO)/fluent/bin/fluentd:32:in `<main>'
> ```

The cause of this error is that the environmental variable has `"`.

```rb
irb(main):001:0> File.read("\"C:/Dummy Program Files (FOO)/fluent/bin/..//etc/fluent/fluentd.conf\"")
(irb):1:in `read': Invalid argument @ rb_sysopen - "C:/Dummy Program Files (FOO)/fluent/bin/..//etc/fluent/fluentd.conf" (Errno::EINVAL)
        from (irb):1:in `<main>'
        from C:/Ruby32-x64/lib/ruby/gems/3.2.0/gems/irb-1.7.4/exe/irb:9:in `<top (required)>'
        from C:/Ruby32-x64/bin/irb:25:in `load'
        from C:/Ruby32-x64/bin/irb:25:in `<main>'
```

kenhys:
Thanks! I'll check it.

kenhys:
[d9b8999](https://github.com/fluent/fluent-package-builder/pull/589/commits/d9b8999dabb2fcd9e3df6ea3b3228f45575ff172) causes msi rollback...

daipom:
Isn't it caused by not applying this?

https://github.com/fluent/fluent-package-builder/pull/589#discussion_r1335325513

kenhys:
> Isn't it caused by not applying this?

maybe that's it. Thanks.

kenhys:
With https://github.com/fluent/fluent-package-builder/commit/e26df3fced2606fb60edc3fc7c35f1d4de338f4a, surely unexpected rollback was fixed.

image

daipom:
What do you think about https://github.com/fluent/fluent-package-builder/pull/589#discussion_r1335315312?

kenhys:
It seems work as expected.

daipom:
Thanks! I will check the built package briefly, just in case.

daipom (approved):
I have confirmed that the build package works as expected! Thanks!

Could you update the title and initial comment?
About the title, since this fix is not only for paths containing space, `which contains space or brackets` or something like that would be better, I think.

kenhys:
Thanks, I'll fix it.

kenhys:
Revised commit message.

daipom:
Thanks!

PR #591, opened 2023-09-28
github: check downgrade to v4

kenhys:
github: check downgrade to v4

    To downgrade correctly you must:

    * Remove fluent-package in beforehand.
    * Manually migrate buffer or logs
    * Re-install td-agent v4.

    NOTE: on debian/ubuntu, need to disable td-agent.service alias
    beforehand because:
      * It conflict with v4 td-agent.service
      * If you didn't disable it, it causes broken symlink from td-agent.service to /lib/systemd/system/fluentd.service.

### DIFF
--- a/fluent-package/apt/systemd-test/downgrade-to-v4.sh
+++ b/fluent-package/apt/systemd-test/downgrade-to-v4.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+set -exu
+
+. $(dirname $0)/../commonvar.sh
+
+# Install v4
+sudo apt install -y curl ca-certificates
+curl -fsSL https://toolbelt.treasuredata.com/sh/install-${distribution}-${code_name}-td-agent4.sh | sh
+
+# td-agent is running
+systemctl status --no-pager td-agent
+
+# fluent-package 5 (LTS)
+#curl -fsSL https://toolbelt.treasuredata.com/sh/install-${distribution}-${code_name}-fluent-package5-lts.sh | sh
+#sudo apt purge -y fluent-lts-apt-source
+
+# Ensure to install the current
+sudo apt install -V -y \
+    /host/${distribution}/pool/${code_name}/${channel}/*/*/fluent-package_*_${architecture}.deb
+
+# td-agent.service is already masked (link to /dev/null), and remove td-agent.service alias not to conflict with v4
+sudo systemctl unmask td-agent
+
+# Even though removing fluent-package, log and .conf are kept. dpkg reports "rc fluent-package" and "rc td-agent" status.
+sudo apt remove -y fluent-package
+
+# fluentd.service is already masked (link to /dev/null), then remove it.
+sudo systemctl unmask fluentd
+
+# Drop symbolic links and recreate real directory.
+sudo rm -f /var/log/td-agent
+sudo rm -f /etc/td-agent
+
+# Migrate logs
+sudo mkdir -p /var/log/td-agent
+sudo chown td-agent:td-agent /var/log/td-agent
+sudo mv /var/log/fluent/* /var/log/td-agent/
+sudo rm -fr /var/log/fluent
+
+# Reinstall v4, but /etc/td-agent is empty.
+sudo mkdir -p /etc/td-agent
+sudo chown td-agent:td-agent /etc/td-agent
+sudo mv /etc/fluent/td-agent.conf /etc/td-agent/
+curl -fsSL https://toolbelt.treasuredata.com/sh/install-${distribution}-${code_name}-td-agent4.sh | sh
+
+# Test: service status
+(! systemctl status --no-pager fluentd)
+(systemctl status --no-pager td-agent)
+
+# Test: config migration
+test -d /etc/td-agent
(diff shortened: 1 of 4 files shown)


daipom:
Thanks!
We need to add this new test here to run it in CI.

https://github.com/fluent/fluent-package-builder/blob/b4b7017c09d78f2c404db62612733b40f69668a6/fluent-package/yum/systemd-test/test.sh#L21-L25

https://github.com/fluent/fluent-package-builder/blob/b4b7017c09d78f2c404db62612733b40f69668a6/fluent-package/apt/systemd-test/test.sh#L26-L30

kenhys:
checking failure on local vm.

kenhys:
Technically package downgrade will work but, not same as before.
Thus, uninstall && re install v4 is the recommended action to do now.

daipom:
Thanks! I'm concerned about the following points.
I didn't think this step was necessary.
I'm checking the behavior.

>     * Disable td-agent.service alias beforehand.
>       
>       * It conflict with v4 td-agent.service
>       * If you didn't disable it, it causes broken symlink from td-agent.service to /lib/systemd/system/fluentd.service.

daipom:
In my local env Rocky Linux 8, I don't need to manually remove the td-agent.service alias.
It is automatically removed when uninstalling fluent-package.

I'm testing https://github.com/daipom/fluent-package-builder/commit/3471b31e036f420e0ed3cd4ff0f9eaf914d2e744 on my fork:

* rpm: https://github.com/daipom/fluent-package-builder/actions/runs/7164750480
* apt: https://github.com/daipom/fluent-package-builder/actions/runs/7164750483

daipom:
> In my local env Rocky Linux 8, I don't need to manually remove the td-agent.service alias. It is automatically removed when uninstalling fluent-package.
> 
> I'm testing [daipom@3471b31](https://github.com/daipom/fluent-package-builder/commit/3471b31e036f420e0ed3cd4ff0f9eaf914d2e744) on my fork:
> 
>     * rpm: https://github.com/daipom/fluent-package-builder/actions/runs/7164750480
> 
>     * apt: https://github.com/daipom/fluent-package-builder/actions/runs/7164750483

Let me sort out the necessary steps for downgrading while looking at these results.

fluent-package/yum/systemd-test/downgrade-to-v4.sh
```diff
@@ -0,0 +1,104 @@
...
+sudo systemctl disable td-agent
+
+sudo $DNF remove -y fluent-package
+
+sudo rm -f /var/log/td-agent
+sudo rm -f /etc/td-agent
+sudo mkdir -p /var/log/td-agent
+sudo chown td-agent:td-agent /var/log/td-agent
+
+sudo mv /var/log/fluent/* /var/log/td-agent/
```
daipom, line 42: For rpm packages, the symlinks for compatibility are automatically removed.
So, we don't need to remove them.

```suggestion
sudo $DNF remove -y fluent-package
(! test -e /etc/td-agent)
(! test -e /var/log/td-agent)
sudo mv /var/log/fluent/ /var/log/td-agent/
```

I have confirmed this works correctly on my fork CI (https://github.com/daipom/fluent-package-builder/actions/runs/7164750480)

fluent-package/apt/systemd-test/downgrade-to-v4.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# Ensure to install the current
+sudo apt install -V -y \
+    /host/${distribution}/pool/${code_name}/${channel}/*/*/fluent-package_*_${architecture}.deb
+
+# td-agent.service is already masked (link to /dev/null), and remove td-agent.service alias not to conflict with v4
+sudo systemctl unmask td-agent
+
+# Keep log and .conf and so on. dpkg reports "rc fluent-package" and "rc td-agent" status.
+sudo apt remove -y fluent-package td-agent
```
daipom, line 26: Do we need to specify `td-agent`?
Wouldn't it be automatically removed?

```suggestion
sudo apt remove -y fluent-package
```

daipom:
> To downgrade correctly you must:
> 
>     * Disable td-agent.service alias beforehand.
>       
>       * It conflict with v4 td-agent.service
>       * If you didn't disable it, it causes broken symlink from td-agent.service to /lib/systemd/system/fluentd.service.
> 
>     * Purge v5 apt-line in fluent-lts-apt-source or fluent-apt-source.

In my conclusion, these are required only for deb.
For rpm, we don't need these steps because symlinks are automatically removed when uninstalling.

fluent-package/yum/systemd-test/downgrade-to-v4.sh
kenhys: Surely no need to remove symlink in beforehand. Thanks.

fluent-package/apt/systemd-test/downgrade-to-v4.sh
kenhys: No need to specify it. Thanks.

daipom (approved):
Thanks! LGTM!

PR #449, opened 2023-04-26
The package name (td-agent) is changed to fluent-package

kenhys:
The package name (td-agent) is changed to fluent-package.

This PR is follow-up of #448 .

NOTE:
* The internal path (such as /opt/td-agent) is not changed.
* Install/Upgrade package test is now testing
  from td-agent 4 to fluent-package 5.
* It provides transitional td-agent package intentionally..
  It should be removed after a while.

It keeps compatibility.

PR #592, opened 2023-10-13
Fix system user/group name in logrotate config

rahulbpatil:
Fix system user/group name in logrotate config

I had installed  `fluent-package` - `5.0.1-1` on ubuntu 22.04
```
ii  fluent-package 5.0.1-1      amd64        All in one package of Fluentd
```

With fluent-package v5 user/group name is changed to `_fluentd`
- https://github.com/fluent/fluent-package-builder/blob/v5.0.1/fluent-package/templates/package-scripts/fluent-package/deb/postinst#L13
- https://github.com/fluent/fluent-package-builder/blob/v5.0.1/fluent-package/templates/package-scripts/fluent-package/deb/postinst#L19

However user/group name in logrotate config is `fluentd`, which causes following failure
```
$ logrotate -f /etc/logrotate.d/fluentd
error: /etc/logrotate.d/fluentd:7 unknown user 'fluentd'
error: found error in /var/log/fluent/fluentd.log , skipping
```
I guess this change was introduced in - https://github.com/fluent/fluent-package-builder/pull/449

### DIFF
--- a/fluent-package/templates/etc/logrotate.d/fluentd
+++ b/fluent-package/templates/etc/logrotate.d/fluentd
@@ -4,7 +4,7 @@
   compress
   delaycompress
   notifempty
-  create 640 <%= service_name %> <%= service_name %>
+  create 640 _<%= service_name %> _<%= service_name %>
   sharedscripts
   postrotate
     pid=/var/run/<%= package_dir %>/<%= service_name %>.pid


daipom:
Thanks. We will confirm this.

Could you add DCO to the commit?
https://github.com/fluent/fluent-package-builder/pull/592/checks?check_run_id=17666702181

daipom:
I have confirmed I can't logrotate by `/etc/logrotate.d/fluent` in fluent-package v5 for deb.
We need to fix this.
Thanks for the report, and sorry for the inconvenience.

I think we need some additional fixes.
In the rpm package, the user and group name are just `fluentd`, not `_fluentd`.
So, we need to fix this without breaking the rpm package.

rahulbpatil:
Added DCO with https://github.com/fluent/fluent-package-builder/pull/592/commits/b5503d1d37497fdfaa3c59b2c388cac8f573d334
- https://github.com/fluent/fluent-package-builder/pull/592/checks?check_run_id=17668371710

kenhys (approved):
LGTM. Thanks!

kenhys:
It seems that some testing CI fails, but it should be fixed in another PR.

daipom:
> I have confirmed I can't logrotate by `/etc/logrotate.d/fluent` in fluent-package v5 for deb. We need to fix this. Thanks for the report, and sorry for the inconvenience.
> 
> I think we need some additional fixes. In the rpm package, the user and group name are just `fluentd`, not `_fluentd`. So, we need to fix this without breaking the rpm package.

TODO: We need to fix this problem.

PR #594, opened 2023-10-16
Fix system user/group name in logrotate.d/fluentd again

kenhys:
This is follow-up of #592 to fix problem correctly.

fluent-package/tempaltes/etc/logrotate.d/fluentd is used
for deb and rpm, so it should support both of different
system user/group respectively.

### DIFF
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -475,25 +475,14 @@ class BuildTask
         end
       end
 
-      desc "Create fluent-package configuration files from template"
-      task :fluent_package_config do
-        fluentd_conf = "etc/#{PACKAGE_DIR}/#{SERVICE_NAME}.conf"
-        fluentd_conf_default = "opt/#{PACKAGE_DIR}/share/#{SERVICE_NAME}.conf"
-        configs = [fluentd_conf]
-        configs.concat([
-          "etc/logrotate.d/#{SERVICE_NAME}",
-          fluentd_conf_default,
-        ]) unless windows? || macos?
-        configs.each do |config|
-          src =
-            if config == fluentd_conf_default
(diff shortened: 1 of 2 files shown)


kenhys:
Extracted task which depends package type.

daipom:
Sorry for my late response.
I'm seeing this now.

fluent-package/Rakefile
```diff
@@ -596,16 +585,16 @@ class BuildTask
...
-      task :deb_config => [:fluent_package_config, :systemd_tmpfiles_config, :bin_scripts, :deb_systemd, :deb_scripts]
+      task :deb_config => [:deb_fluent_package_config, :systemd_tmpfiles_config, :bin_scripts, :deb_systemd, :deb_scripts]
 
       desc "Create configuration files for Windows"
-      task :msi_config => [:fluent_package_config, :wix_config, :win_batch_files]
+      task :msi_config => [:rpm_fluent_package_config, :wix_config, :win_batch_files]
 
       desc "Create configuration files for macOS"
-      task :dmg_config => [:fluent_package_config, :pkgbuild_scripts, :pkgbuild_config, :bin_scripts, :launchctl_config]
+      task :dmg_config => [:rpm_fluent_package_config, :pkgbuild_scripts, :pkgbuild_config, :bin_scripts, :launchctl_config]
```
daipom, line 597: For future tasks, we need to improve these codes by separating the role of the `render_fluent_package_config` method.
First of all, I think there should be variables for the user and group.
(I think it should be done in another PR.)

daipom (approved):
LGTM. Thanks.
I have confirmed that I can rotate the log file in the package for Ubuntu Focal.

PR #596, opened 2023-10-31
rpm deb: add user as system user

kenhys:
When upgrading from v4, there is a possibility that fluentd user is not  created as system user.

Before:
      Without --system, /var/lib/fluent will be created

 After:
      With --system, /var/lib/fluent will NOT be created

NOTE: Not to overlook it, use long option explicitly.

UID itself is kept in range of system user (td-agent), so not accessed via normal user.


daipom (approved):
LGTM. Thanks!

PR #598, opened 2023-11-07
 rpm: do not delete system account

kenhys:
In the previous versions (5.0.0, 5.0.1), when fluent-package was
removed, system account (user and group are intended to be removed)

But there is a case that it fails to remove fluentd group because of
compatible GID is assigned for td-agent and fluentd when
fluent-package was introduced with upgrading from td-agent v4.

  Removing fluentd user...
  userdel: group fluentd is the primary group of another user and is not removed.
  Removing fluentd group...
  groupdel: cannot remove the primary group of user 'td-agent'
  Removing td-agent user...
  userdel: td-agent mail spool (/var/spool/mail/td-agent) not found
  userdel: td-agent home directory (/var/lib/td-agent) not found

This is a bug of fluent-package apparently.

This kind of inconsistency causes maintainer script error when
reinstalling td-agent or fluent-package again.

And moreover, if system account (user and group) was removed
completely, there is a case that no user can access generated logs
when user re-installed td-agent or fluent-package. (mismatch of
UID/GID which is newly created)

This case also should be considered.
(Keep system account after removing package)

### DIFF
--- a/fluent-package/yum/install-test.sh
+++ b/fluent-package/yum/install-test.sh
@@ -83,8 +83,8 @@ for conf_path in /etc/td-agent/td-agent.conf /etc/fluent/fluentd.conf; do
     fi
 done
 
-(! getent passwd fluentd >/dev/null)
-(! getent group fluentd >/dev/null)
+getent passwd fluentd >/dev/null
+getent group fluentd >/dev/null
 
 if [ $ENABLE_UPGRADE_TEST -eq 1 ]; then
     echo "UPGRADE TEST from v4"
--- a/fluent-package/yum/systemd-test/install-newly.sh
+++ b/fluent-package/yum/systemd-test/install-newly.sh
@@ -46,9 +46,18 @@ test -e /var/log/fluent/fluentd.log
 sudo $DNF remove -y fluent-package
 sudo systemctl daemon-reload
 
-getent passwd fluentd >/dev/null
-getent group fluentd >/dev/null
-
+case $1 in
+  local)
+    getent passwd fluentd >/dev/null
+    getent group fluentd >/dev/null
+    ;;
+  *)
+    # TODO: Remove this branch after the following fix is applied to the latest release.
+    # https://github.com/fluent/fluent-package-builder/pull/598
+    (! getent passwd fluentd >/dev/null)
+    (! getent group fluentd >/dev/null)
+    ;;
+esac
 # `sudo systemctl daemon-reload` clears the service completely.
 #   (The result of `systemctl status` will be `unfound`)
 # Note: RPM does not leave links like `@/etc/systemd/system/fluentd.service`.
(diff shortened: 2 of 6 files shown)


fluent-package/yum/install-test.sh
```diff
@@ -83,8 +83,8 @@ for conf_path in /etc/td-agent/td-agent.conf /etc/fluent/fluentd.conf; do
     fi
 done
 
-(! getent passwd fluentd >/dev/null)
-(! getent group fluentd >/dev/null)
+getent passwd fluentd >/dev/null
+getent group fluentd >/dev/null
```
daipom, line 87: This fails for the `v5` and `lts` tests until the v5.0.2 is released.
How about this?

```suggestion
case $1 in
  local)
    getent passwd fluentd >/dev/null
    getent group fluentd >/dev/null
    ;;
  *)
    # TODO: Remove this branch after the following fix is applied to the latest release.
    # https://github.com/fluent/fluent-package-builder/pull/598
    (! getent passwd fluentd >/dev/null)
    (! getent group fluentd >/dev/null)
    ;;
esac
```

kenhys: Wait, I'll revert it.

kenhys: yum/install-test.sh always tests locally build package.
Thus, it doesn't matter that is released or not.

fluent-package/yum/systemd-test/install-newly.sh
```diff
@@ -46,6 +46,9 @@ test -e /var/log/fluent/fluentd.log
 sudo $DNF remove -y fluent-package
 sudo systemctl daemon-reload
 
+getent passwd fluentd >/dev/null
+getent group fluentd >/dev/null
+
```
daipom, line 51: I'm sorry.
I was suggesting the wrong file.
It should be `yum/systemd-test/install-newly.sh`, not `yun/install-test.sh`.

```suggestion
case $1 in
  local)
    getent passwd fluentd >/dev/null
    getent group fluentd >/dev/null
    ;;
  *)
    # TODO: Remove this branch after the following fix is applied to the latest release.
    # https://github.com/fluent/fluent-package-builder/pull/598
    (! getent passwd fluentd >/dev/null)
    (! getent group fluentd >/dev/null)
    ;;
esac
```

fluent-package/yum/install-test.sh
daipom: I'm sorry.
I was suggesting the wrong file.

fluent-package/yum/systemd-test/install-newly.sh
kenhys: Ah, got it. Thanks!

daipom (approved):
LGTM. Thanks!

PR #600, opened 2023-11-22
td-agent v4.5.2

ashie:


ashie:
The following issue is considered as fatal in certain usage, so we are planning to release td-agent v4.5.2.
https://github.com/fluent/fluentd/pull/4327

daipom (approved):
LGTM. Thanks!

PR #606, opened 2023-12-12
msi: fix wrong env path for Fluent Package Prompt

daipom:
* fix https://github.com/fluent/fluent-package-builder/issues/605
* The current PATH includes `"`. It is unintended and causes some
  bugs. For example, fluent-diagtool does not work correctly
  because of this.

PR #607, opened 2023-12-13
msi: remove unnecessary path delimiter

daipom:
The last character of `%~dp0` is `\`, so we don't need add it.
(Though it appears it does not cause any problems.)


daipom:
Note: https://github.com/fluent/fluent-package-builder/pull/606#issuecomment-1851432074

kenhys (approved):
LGTM.

Thanks.

PR #613, opened 2024-02-05
rpm: take over enabled state of systemd service from v4

kenhys:
In the previous versions, the enabled service status of td-agent was not migrated.

This commit try to take over existing td-agent service status.

Note that if systemctl enable fluentd is executed in %post section, it was reverted in td-agent's %preun scriptlet.
So it is enabled in %posttrans explicitly.

Closes: #610

PR #615, opened 2024-02-09
deb rpm: quote target files

kenhys:
There was a case that missing quote about file name causes migration failures.

* No quote under /etc/td-agent/'PATH_WITH_SPACE'
* No quote under /var/log/td-agent/'PATH_WITH_SPACE'

Usually such files should not be generated, but need to care it in practical use case.

### DIFF
--- a/fluent-package/apt/systemd-test/update-from-v4.sh
+++ b/fluent-package/apt/systemd-test/update-from-v4.sh
@@ -19,6 +19,14 @@ sudo apt install -V -y td-agent=${td_agent_version}-1
 
 systemctl status --no-pager td-agent
 
+# Generate garbage files
+touch /etc/td-agent/a\ b\ c
+touch /var/log/td-agent/a\ b\ c.log
+touch /etc/td-agent/plugin/in_fake.rb
+for d in $(seq 1 10); do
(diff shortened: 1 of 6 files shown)


daipom:
Thanks!

Are there file names that are not covered by this?

```bash
mv -f '/etc/<%= compat_package_dir %>/$d' '/etc/<%= package_dir %>/'
```

kenhys:
> mv -f '/etc/<%= compat_package_dir %>/$d' '/etc/<%= package_dir %>/'

It is not covered because of combination with `for d in ...` doesn't work as expected.

daipom:
Oh, sorry. As you say, we can't use `'` here.
What about `"`?

```
mv -f "/etc/<%= compat_package_dir %>/$d" '/etc/<%= package_dir %>/'
```

kenhys:
As #613 was merged, rebased with master.

kenhys:
> This looks good, but is there any reason you didn't use for d in $(ls /etc/<%= compat_package_dir %>)?

I've just  overlooked that `for d in  "$(ls /etc/<%= compat_package_dir %>/)"` would work.

kenhys:
Thank you for reviewing!

Issue #611, opened 2024-01-24
Prevent accidental duplicate launching


daipom:
`fluentd` or `td-agent` command launches Fluentd with the default config path.
This makes it easy to start Fluentd, but it also causes accidental duplicate launching.

For example, it launches Fluentd duplicated while Fluentd is running as a systemd unit.

Processing the same files, such as buffer files, pos files, and output destination files, causes unassumed errors.
At worst, it may cause file corruption or data loss.

The current specification makes it too easy to start Fluentd, so I'm concerned about mishandling.
For example, trying to check the version and launching Fluentd by mistake, etc.

# The current specification

 The default config path is set in env variables.

https://github.com/fluent/fluent-package-builder/blob/c75754e6619dab73920fdb386c0de67b77c83e22/fluent-package/templates/usr/sbin/fluentd.erb#L4

So, we can start Fluentd without any command arguments.

fluent-package:

```console
$ fluentd
```

td-agent:

```console
$ td-agent
```

# What I want to do

I want to add some prevention.
For example,

* Add confirmation that you are really trying to start Fluentd

It would change the specifications, so we need to be careful about compatibility.

kenhys:
It may be better to implement such a guard feature in fluentd not fluent-package-builder.

kenhys:
https://github.com/fluent/fluentd/blob/master/lib/fluent/command/fluentd.rb#L343
It is enough to guard before starting service appropriately. :thinking:

daipom:
> It may be better to implement such a guard feature in fluentd not fluent-package-builder.

I too think it may end up that way.
However, since the specification of easy duplicate launching is only a problem of the package, it is necessary to consider how to take appropriate countermeasures on either side.

kenhys:
Use case on Windows:

* Already started fluentd as a service (fluentdwinsvc)
* Launch Fluent Package Prompt with Administrator privilege
* Execute another instance via fluentd -v 
  
If fluentdopt is not changed, usually c:/opt/fluent/etc/fluent/fluentd.conf will be used  in both cases.
Thus, blocking from c:/opt/fluent/fluentd.bat is effective if it will be fixed in fluent-package side.

```
c:\opt\fluent>where fluentd
c:\opt\fluent\fluentd.bat
c:\opt\fluent\bin\fluentd
c:\opt\fluent\bin\fluentd.bat
```

Use case in Linux:

* Already started fluentd as a service (fluentdwinsvc)
* Execute another instance via fluentd -v

[...]

PR #617, opened 2024-02-15
linux: guard launching duplicated fluentd instance with same configuration

kenhys:
linux: guard duplicated instance (not adding a new option)

Closes: #611 

Before:

We can launch Fluentd by /usr/sbin/fluentd even though fluentd service
is running.

Launching multiple Fluentd with the same config may cause
inconsistency of the buffers or the pos files.

After:

We can't launch Fluentd by /usr/sbin/fluentd with the default config
path if fluentd service is running.  If one of the following options is
specified, we can execute fluentd as before.

* --config (-c)
* --dry-run

```
$ sudo /usr/sbin/fluentd 
Error: Can't start duplicate Fluentd instance with the default config.
To start Fluentd, please do one of the following:
(Caution: Please be careful not to start multiple instances with the same config.)
- Stop the Fluentd service 'fluentd'.
- Specify the config path explicitly by '-c' ('--config').
```

### DIFF
--- a/fluent-package/templates/usr/sbin/fluentd.erb
+++ b/fluent-package/templates/usr/sbin/fluentd.erb
@@ -16,7 +16,7 @@ end
 if RUBY_PLATFORM =~ /linux/
   prevent_duplicate_launch = system("systemctl", "is-active", "fluentd", out: IO::NULL)
   if prevent_duplicate_launch
-    unless ARGV.include?("-c") or ARGV.include?("--config") or ARGV.include?("--dry-run")
+    if ["-c", "--config", "--dry-run"].none? {|allowing_opt| ARGV.include? allowing_opt}
       puts("Error: Can't start duplicate Fluentd instance with the default config.")
       if ARGV.include?("-v")
         puts("To take the version, please use '--version', not '-v' ('--verbose').")
--- a/fluent-package/templates/usr/sbin/fluentd.erb
+++ b/fluent-package/templates/usr/sbin/fluentd.erb
@@ -19,7 +19,7 @@ if RUBY_PLATFORM =~ /linux/
     unless ARGV.include?("-c") or ARGV.include?("--config") or ARGV.include?("--dry-run")
       puts("Error: Can't start duplicate Fluentd instance with the default config.")
       if ARGV.include?("-v")
-	puts("To take the version, please use '--version', not '-v' ('--verbose').")
+        puts("To take the version, please use '--version', not '-v' ('--verbose').")
       end
       puts <<EOS
 To start Fluentd, please do one of the following:
--- a/fluent-package/templates/usr/sbin/fluentd.erb
+++ b/fluent-package/templates/usr/sbin/fluentd.erb
@@ -12,4 +12,38 @@ if ARGV.include?("--version")
   puts "fluent-package #{PACKAGE_VERSION} fluentd #{Fluent::VERSION} (#{FLUENTD_REVISION})"
   exit 0
 end
+
+class DuplicateInstanceChecker
+  def running_fluentd?
+    `systemctl is-active fluentd`
+    return true if $?.exitstatus == 0
+    false
+  end
+
+  def prevent_duplicate_launch?
+    unless running_fluentd?
+      return false
+    end
+    true
+  end
+end
+
+if RUBY_PLATFORM =~ /linux/
+  checker = DuplicateInstanceChecker.new
+  if checker.prevent_duplicate_launch?
+    unless ARGV.include?("-c") or ARGV.include?("--config") or ARGV.include?("--dry-run")
+      puts("Error: Can't start duplicate Fluentd instance with the default config.")
+      if ARGV.include?("-v")
+	puts("To take the version, please use '--version', not '-v' ('--verbose').")
+      end
+      puts <<EOS
+To start Fluentd, please do one of the following:
+(Caution: Please be careful not to start multiple instances with the same config.)
+- Stop the Fluentd service 'fluentd'.
+- Specify the config path explicitly by '-c' ('--config').
+EOS
+      exit 2
+    end
+  end
+end
 load "<%= install_path %>/bin/fluentd"
--- a/fluent-package/templates/usr/sbin/fluentd.erb
+++ b/fluent-package/templates/usr/sbin/fluentd.erb
@@ -13,24 +13,9 @@ if ARGV.include?("--version")
   exit 0
 end
 
-class DuplicateInstanceChecker
-  def running_fluentd?
-    `systemctl is-active fluentd`
-    return true if $?.exitstatus == 0
-    false
-  end
-
-  def prevent_duplicate_launch?
-    unless running_fluentd?
-      return false
-    end
-    true
-  end
-end
-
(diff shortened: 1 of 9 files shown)


kenhys:
NOTE: /opt/fluent/bin/fluentd  is auto-generated file, so it should be patched.

kenhys:
Observed service restarting in short times.
Need to investigating further more.

kenhys:
Update comment in this PR.

kenhys:
Unexpectedly, download-artifact fails.

```
actions/download-artifact@master
Downloading single artifact
Preparing to download the following artifacts:
- packages-rockylinux-8 (ID: 1274111102, Size: 108455247)
Redirecting to blob download url: https://productionresultssa8.blob.core.windows.net/actions-results/b9412b87-4106-46a2-8fb0-86c757be7f85/workflow-job-run-ff7a8386-1a15-5f2b-97b2-c907b43b2f17/artifacts/6d0d09c616a6cee6f67320471aa47616aa04bd135f06b78e56dad6b651114b8f.zip
Starting download of artifact to: /home/runner/work/fluent-package-builder/fluent-package-builder
(node:1729) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Error: Unable to download artifact(s): Unable to download and extract artifact: Artifact download failed after 5 retries.
```

kenhys:
It seems that there is no need to stick to download-artifact@master, created PR to use stable download-artifact@v4.
https://github.com/fluent/fluent-package-builder/pull/624

kenhys:
NOTE:
download failure issue is not related to whether using download-artifact@master.

 
[bug] Node incompatibility: [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues 
https://github.com/actions/download-artifact/issues/283

kenhys:
got it, it may be missing runtime package dependency.

kenhys:
checking with timeout.

kenhys:
As v5.0.3 is not released yet, test case in fresh install v5, lts should be skipped.

kenhys:
With debug commit, it has passed on bookworm 

https://github.com/fluent/fluent-package-builder/actions/runs/8077515347/job/22068761536?pr=617

image

On Rocky

https://github.com/fluent/fluent-package-builder/actions/runs/8077515348/job/22068663355

image

kenhys:
It should pass CI. waiting results.

kenhys:
Failing on windows, but not related with this PR.

https://github.com/fluent/fluent-package-builder/actions/runs/8077907378/job/22069129797?pr=617
something wrong with ffi gem dependency.

kenhys:
> https://github.com/fluent/fluent-package-builder/actions/runs/8077907378/job/22069129797?pr=617
something wrong with ffi gem dependency.

It may be not accurate.

https://github.com/fluent/fluent-package-builder/actions/runs/8076317129/job/22064585573
The version of ffi related gems are same as before.

kenhys:
> Failing on windows, but not related with this PR.

It is a bit strange, but it is not reproducible now...

kenhys:
All checks has passed now!

kenhys:
Make it simple solution:

A) guard /usr/sbin/fluentd and /opt/fluent/bin/fluentd
* need to inject specific code to block  duplicated instance.

B) guard only /usr/sbin/fluentd
* it may be simpler, but can't block via /opt/fluent/bin/fluentd.

kenhys:
waiting CI to pass...

fluent-package/templates/usr/sbin/fluentd.erb
```diff
@@ -12,4 +12,52 @@ if ARGV.include?("--version")
...
+        # worker
+	return false
+      else
+        return true
+      end
+    rescue
+      return true
+    end
+    false
+  end
```
daipom, line 42: Thanks for this fix!
I'm concerned about this.
Does this mean `/usr/sbin/fluentd` cab be called from the supervisor process?
Can we reproduce this by just restarting the service?

kenhys: No. 
This issue is raised when `/usr/sbin/fluentd` was invoked.
(service file use /opt/fluent/bin/fluentd)

fluent-package/templates/usr/sbin/fluentd.erb
```diff
@@ -12,4 +12,38 @@ if ARGV.include?("--version")
...
+    unless running_fluentd?
+      return false
+    end
+    true
+  end
+end
+
+if RUBY_PLATFORM =~ /linux/
+  checker = DuplicateInstanceChecker.new
+  if checker.prevent_duplicate_launch?
```
daipom, line 33: How about simply this?

```suggestion
if RUBY_PLATFORM =~ /linux/
  prevent_duplicate_launch = system("systemctl", "is-active", "fluentd", out: IO::NULL)
  if prevent_duplicate_launch
```

fluent-package/templates/usr/sbin/fluentd.erb
daipom: Note: We don't need this check for workers, this check has been removed.

fluent-package/templates/usr/sbin/fluentd.erb
```diff
@@ -12,4 +12,38 @@ if ARGV.include?("--version")
...
+    unless ARGV.include?("-c") or ARGV.include?("--config") or ARGV.include?("--dry-run")
+      puts("Error: Can't start duplicate Fluentd instance with the default config.")
+      if ARGV.include?("-v")
+	puts("To take the version, please use '--version', not '-v' ('--verbose').")
```
daipom, line 37: 1 tab to 8 spaces

```suggestion
        puts("To take the version, please use '--version', not '-v' ('--verbose').")
```

fluent-package/templates/usr/sbin/fluentd.erb
kenhys: Thanks, it looks good!

fluent-package/templates/usr/sbin/fluentd.erb
```diff
@@ -12,4 +12,23 @@ if ARGV.include?("--version")
   puts "fluent-package #{PACKAGE_VERSION} fluentd #{Fluent::VERSION} (#{FLUENTD_REVISION})"
   exit 0
 end
+
+if RUBY_PLATFORM =~ /linux/
+  prevent_duplicate_launch = system("systemctl", "is-active", "fluentd", out: IO::NULL)
+  if prevent_duplicate_launch
+    unless ARGV.include?("-c") or ARGV.include?("--config") or ARGV.include?("--dry-run")
```
daipom, line 19: This is a minor point, but how about this?

```suggestion
    if ["-c", "--config", "--dry-run"].none? {|allowing_opt| ARGV.include? allowing_opt}
```

daipom (approved):
I have commented on one minor point, but this basically looks good to me.
Thanks for this!
I have confirmed this script on my local AlmaLinux 9, and it works well!

PR #624, opened 2024-02-26
github: use stable actions/download-artifact@v4

kenhys:
actions/download-artifact@master branch is already deprecated and switched to main branch. 
(master branch is 121 commits behind main)

There is no need to always dare to follow main branch, so use stable v4 release.


daipom (approved):
LGTM. Thanks.

PR #636, opened 2024-03-25
github: extract test file as element of matrix

kenhys:
Before:

  test.sh execute multiple test files.
  This behavior make it hard to distinct failure test cases.

After:

  Use test.sh as launch script. It doesn't execute multiple test files.
  This change was introduced to make it easy to distinct each test
  cases on error.

### DIFF
--- a/fluent-package/apt/systemd-test/test.sh
+++ b/fluent-package/apt/systemd-test/test.sh
@@ -2,50 +2,34 @@
 
 if [ -z $1 ]; then
     echo "Error: Need to specify lxc image name."
-    echo "Ex.) $ ./test.sh ubuntu:20.04"
+    echo "Ex.) $ ./test.sh ubuntu:20.04 "
+    echo "Ex. CI) $ ./test.sh ubuntu:20.04 install-newly.sh local"
     exit 1
 fi
 
 image=$1
+test_file=$2
+apt_repo_type=$3
 dir="/host/fluent-package/apt/systemd-test"
 
 set -eux
 
-test_filenames=(
-    update-to-next-version.sh
-)
-
-if [ ! $image = "images:debian/12" ]; then
-    # As no bookworm package for v4, so execute upgrade test for other code name.
-    test_filenames+=(
-        update-from-v4.sh
-        update-to-next-version-with-backward-compat-for-v4.sh
-        downgrade-to-v4.sh
-    )
-fi
-
-for apt_repo_type in local v5 lts; do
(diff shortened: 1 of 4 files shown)


daipom:
I have one concern.
Wouldn't it make running tests in the local environment difficult?

I don't run tests on my local these days.
This concern may be unnecessary. I have commented on this just in case.

Come to think of it, I was thinking that I could use `grouping log lines` feature to make the results easier to understand before.

* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines

I thought that this could be used as follows:

```sh
for test_filename in ${test_filenames[@]}; do
    echo -e "::group::Run test: $test_filename\n"
    ...
    echo "::endgroup::"
done
```

daipom:
> I have one concern. Wouldn't it make running tests in the local environment difficult?
> 
> I don't run tests on my local these days. This concern may be unnecessary. I have commented on this just in case.

Sorry, I don't think we need to worry too much about this point.
When I run tests locally, I'm likely to run specific tests repeatedly.
It may be enough to use CI to run all tests.

kenhys:
Thanks, 

> Wouldn't it make running tests in the local environment difficult?

It may be. so it takes into account by `$CI`.

kenhys:
> Come to think of it, I was thinking that I could use grouping log lines feature to make the results easier to understand before.

I didn't know such a feature, Thanks!

kenhys:
re-run stalled jobs.

fluent-package/apt/systemd-test/test.sh
```diff
@@ -2,50 +2,73 @@
...
+	lxc launch $image target
+	sleep 5
+	lxc config device add target host disk source=$PWD path=/host
+	lxc list
+	lxc exec target -- $dir/setup.sh
+	lxc exec target -- $dir/$test_filename
+	lxc stop target
+	lxc delete target
+    done
+fi
```
daipom, line 73: I'm sorry if I am confusing you.
If you don't need this logic for the not CI case, then I think we can remove it.
It is because, now, I think as follows,

> Sorry, I don't think we need to worry too much about this point.
When I run tests locally, I'm likely to run specific tests repeatedly.
It may be enough to use CI to run all tests.

and it would be annoying to maintain a list of test files on both the CI side and the script side.

If you need it, no problem.
If it's because of my previous suggestion, we should remove it and simplify the script.

I'm sorry if I am confusing you.

daipom (approved):
LGTM, basically!
I just commented on a little thing.

Issue #618, opened 2024-02-16
Windows: Slow service start (fluent-package)


daipom:
# Abst

* Windows service startup is noticeably slower for fluent-package
* This problem can cause timeout errors (EventID 7000 or EventID 7009)
* This problem occurs in environments with multiple CPUs
* This is the problem of [win32-service](https://github.com/chef/win32-service) with Ruby 3 series being slow
  * https://github.com/chef/win32-service/issues/84

# Details

Windows service startup is noticeably slower for fluent-package.
I'm talking about the time to reach the `Running` state. In other words, the time to finish [StartServiceCtrlDispatcher](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-startservicectrldispatchera).
For td-agent, it takes about 5 seconds, but for fluent-package it can take more than 20 seconds.

Windows service needs to finish [StartServiceCtrlDispatcher](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-startservicectrldispatchera) in 30 seconds by default.
So, this problem can cause timeout errors.

EventID 7000

```
The testservice service failed to start due to the following error: 
The service did not respond to the start or control request in a timely fashion.
```

EventID 7009

```
A timeout was reached (30000 milliseconds) while waiting for the testservice service to connect.
```

# Reproduce

If you try to start the service several times, it will often take 10 to 20 seconds.

I confirm this problem in the following:

* Windows 10 Home 64bit
  * 4 core CPU, 8 logical processors (Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz 1.99GHz)
  * 16GB RAM
* AWS EC2 instance Windows Server 2019 Base
  * t2.medium
  * t2.large
  * c5.large

Interestingly, this problem does not occur in t2.small EC2 instance type.
I think it occurs in environments with multiple CPUs.

daipom:
https://github.com/chef/win32-service/pull/85 will fix this issue.

kenhys:
it should be fixed via #630

PR #637, opened 2024-03-25
CI: Check time to start service on Windows

daipom:
Add a test for

* #618

On v5.0.2, I have confirmed it takes 4.6s ~ 6.7s.

https://github.com/daipom/fluent-package-builder/actions/runs/8417853364/job/23047108305#step:7:168

On the master, it takes less than 1s.

https://github.com/daipom/fluent-package-builder/actions/runs/8417835331/job/23047050065#step:7:167

### DIFF
--- a/fluent-package/msi/install-test.ps1
+++ b/fluent-package/msi/install-test.ps1
@@ -4,14 +4,35 @@ $msi = ((Get-Item "C:\\fluentd\\fluent-package\\msi\\repositories\\fluent-packag
 Write-Host "Installing ${msi} ..."
 
 Start-Process msiexec -ArgumentList "/i", $msi, "/quiet" -Wait -NoNewWindow
-Start-Service fluentdwinsvc
+$initialStatus = (Get-Service fluentdwinsvc).Status
+if ($initialStatus -ne "Stopped") {
+    Write-Host "The initial status must be 'Stopped', but it was '${initialStatus}'."
+    [Environment]::Exit(1)
+}
 
 $ENV:PATH="C:\\opt\\fluent\\bin;" + $ENV:PATH
 $ENV:PATH="C:\\opt\\fluent;" + $ENV:PATH
 
 td-agent --version
 
-Start-Sleep 3
+# https://github.com/fluent/fluent-package-builder/issues/618
+$thresholdSeconds = 6
+Write-Host "Measuring times to start the service..."
+$timeSpans = 0..2 | % {
+    Measure-Command { Start-Service fluentdwinsvc }
+    Start-Sleep 15
+    Stop-Service fluentdwinsvc
+    Start-Sleep 15
+}
+Write-Host "Measured seconds to start the service:"
+$timeSpans | %{ Write-Host $_.TotalSeconds }
+if (($timeSpans | Measure-Object -Property TotalSeconds -Maximum).Maximum -gt $thresholdSeconds) {
+    # It should be 0.5s ~ 3s because starting service should be done immediately.
+    # (The only things that take time are starting Ruby and loading the libraries.)
+    Write-Host "Launching is abnormally slow. (The max value is greater than ${thresholdSeconds}s)"
+    [Environment]::Exit(1)
+}
+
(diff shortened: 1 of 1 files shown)


fluent-package/msi/install-test.ps1
```diff
@@ -4,14 +4,30 @@ $msi = ((Get-Item "C:\\fluentd\\fluent-package\\msi\\repositories\\fluent-packag
...
 
 td-agent --version
 
-Start-Sleep 3
+Write-Host "Measuring times to start the service..."
+$timeSpans = 0..2 | % {
+    Measure-Command { Start-Service fluentdwinsvc }
+    Start-Sleep 15 | Out-Null
+    Stop-Service fluentdwinsvc | Out-Null
+    Start-Sleep 15 | Out-Null
```
kenhys, line 23: I'm not sure `|Out-Null` is required.
I thought no Out-Null affects the result of $timeSpans, but it seems not.


```
Start-Sleep 15
Stop-Service fluentdwinsvc
Start-Sleep 15
```

fluent-package/msi/install-test.ps1
```diff
@@ -4,14 +4,30 @@ $msi = ((Get-Item "C:\\fluentd\\fluent-package\\msi\\repositories\\fluent-packag
...
+}
+if (($timeSpans | Measure-Object -Property TotalSeconds -Maximum).Maximum -gt 10) {
+    Write-Host "Launching is abnormally slow:"
+    $timeSpans | %{ Write-Host $_.TotalSeconds }
```
kenhys, line 27: How about always shows $timeSpans TotalSeconds?  
Surely abnormally slow case is important, but it is meaningful to show measured timestamp `<= 10` seconds.

fluent-package/msi/install-test.ps1
daipom: Oh! Really!
I want this for-each to collect only the results of `Measure-Command`, so I force other commands not to output any results.
However, if it does not affects the results, it may be unnecessary.
I will check it later.

Thanks!

fluent-package/msi/install-test.ps1
daipom: Nice idea! I will fix it!

fluent-package/msi/install-test.ps1
daipom: These commands do not output to the standard output, so we don't need `Out-Null` here!
I have removed them.
Thanks!

fluent-package/msi/install-test.ps1
daipom: I have fixed it. Thanks!

fluent-package/msi/install-test.ps1
```diff
@@ -4,14 +4,34 @@ $msi = ((Get-Item "C:\\fluentd\\fluent-package\\msi\\repositories\\fluent-packag
...
+    Start-Sleep 15
+    Stop-Service fluentdwinsvc
+    Start-Sleep 15
+}
+Write-Host "Measured seconds to start the service:"
+$timeSpans | %{ Write-Host $_.TotalSeconds }
+if (($timeSpans | Measure-Object -Property TotalSeconds -Maximum).Maximum -gt 6) {
+    # It should be 0.5s ~ 3s because starting service should be done immediately.
+    # (The only things that take time are starting Ruby and loading the libraries.)
+    Write-Host "Launching is abnormally slow."
```
kenhys, line 31: It is helpful to show "how slow"  with Write-Host.

e.g.  "Launching is abnormally slow. (>= 6  seconds)"

kenhys (changes requested):
Except error message(it may be trivial change), it looks good.

fluent-package/msi/install-test.ps1
daipom: I see! Fixed! Thanks!

daipom:
Applied suggestions and rebased to the latest master.
Checking CI...

kenhys (approved):
LGTM.

daipom:
Thanks!

PR #641, opened 2024-03-29
msi: stop migration run every update

### LINKED ISSUE #640: fluent-package: msi: migration process wrongly runs every update in env once updated from v4
The migration process `fluent-package-post-migration.bat` wrongly runs every update in env once updated from v4.

https://github.com/fluent/fluent-package-builder/blob/9d718761b04a1b028ee0925727f7e7aeaabb67ac/fluent-package/msi/assets/fluent-package-post-migration.bat

This is because the following `if` always becomes true in an environment once updated from v4.

https://github.com/fluent/fluent-package-builder/blob/9d718761b04a1b028ee0925727f7e7aeaabb67ac/fluent-package/msi/assets/fluent-package-post-migration.bat#L4

I have confirmed the following overwriting `fluentd.conf` can be harmful. 

https://github.com/fluent/fluent-package-builder/blob/9d718761b04a1b028ee0925727f7e7aeaabb67ac/fluent-package/msi/assets/fluent-package-post-migration.bat#L25

If adding some setting to `fluentd.conf` and not editing the old config `td-agent.conf`, this overwrites `fluentd.conf` by the old config `td-agent.conf`.
This results in the loss of the current config.

The workaround is to remove the old config file `td-agent.conf`.
We should surely remove this old file before updating v5 from v5. (If that env was once updated from v4.)

daipom:
* Fix #640

### DIFF
--- a/fluent-package/msi/update-from-v5-test.ps1
+++ b/fluent-package/msi/update-from-v5-test.ps1
@@ -47,7 +47,7 @@ $outputFiles = Get-ChildItem "C:\\opt\\td-agent\\output"
 Start-Sleep 5
 $outputFilesAfterSleep = Get-ChildItem "C:\\opt\\td-agent\\output"
 if ($outputFilesAfterSleep.Count -le $outputFiles.Count) {
-    Write-Error "The previous config does not work. Output file num: ${$outputFiles.Count} to ${$outputFilesAfterSleep.Count}."
+    Write-Error ("The previous config does not work. Output file num: {0} to {1}." -f $outputFiles.Count, $outputFilesAfterSleep.Count)
 }
 
 # TODO: Add tests here when it becomes able to take over the fluentdopt and autostart configuration.
--- a/fluent-package/msi/update-from-v5-test.ps1
+++ b/fluent-package/msi/update-from-v5-test.ps1
@@ -1,0 +1,53 @@
+$ErrorActionPreference = "Stop"
+$ProgressPreference = "SilentlyContinue"
+Set-PSDebug -Trace 1
+
+$v4MsiFilename = "td-agent-4.5.3-x64.msi"
+$v5MsiFilename = "fluent-package-5.0.2-x64.msi"
+
+Invoke-WebRequest "https://s3.amazonaws.com/packages.treasuredata.com/4/windows/$v4MsiFilename" -OutFile $v4MsiFilename
+Invoke-WebRequest "https://s3.amazonaws.com/packages.treasuredata.com/5/windows/$v5MsiFilename" -OutFile $v5MsiFilename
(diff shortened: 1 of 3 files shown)


daipom:
I have separated the commits for visibility.

The actual fix is 9ce35a58f33868597cc2160b54a480eb29aa4399.

daipom:
TODO: want to make a test for this.

daipom:
I have added a test.

Now, I will check the reasonability of the test and consider another way to fix this that does not depend on PowerShell.

daipom:
> consider another way to fix this that does not depend on PowerShell.

Done!
The actual fix is f29fa581837d5af7107db4dc7f638305b00cfd17.

daipom:
TODO: check the reasonability of the test

fluent-package/msi/update-from-v5-test.ps1
```diff
@@ -0,0 +1,53 @@
...
+Start-Sleep 30 # Must wait until all processes are surely started.
+
+# Test: The previous config works as before
+$outputFiles = Get-ChildItem "C:\\opt\\td-agent\\output"
+Start-Sleep 5
+$outputFilesAfterSleep = Get-ChildItem "C:\\opt\\td-agent\\output"
+if ($outputFilesAfterSleep.Count -le $outputFiles.Count) {
+    Write-Error "The previous config does not work. Output file num: ${$outputFiles.Count} to ${$outputFilesAfterSleep.Count}."
+}
+
```
kenhys, line 52: It needs 2 test cases.

* v4 => v5.0.2 
* v4 => v5.0.2 => v5.0.x 

In both cases, it may be straight forward to check "not copied into"
I'm not sure that count is enough or not.

daipom: > v4 => v5.0.2

Isn't this done by `update-from-v4-test`?

> In both cases, it may be straight forward to check "not copied into"
> I'm not sure that count is enough or not.

If the current config file is overwritten, no file will be output.
Do you have a concern about this?

daipom: It may be enough to verify that at least one file exists.

kenhys: Oops, I've misunderstand behavior.

kenhys:
I've pushed a commit to fix format issue.

daipom:
> I've pushed a commit to fix format issue.

Oh, thanks!

daipom:
> TODO: check the reasonability of the test

Done.
I will check the built package behavior just in case, but I think this PR can be merged now.

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #643, opened 2024-04-10
ci deb: extract confluent test case

kenhys:
In the previous versions, testing with confluent is
executed in serverspec:linux.

As confluent test case should be out of scope about basic serverspec
to distinct confluent's failure, so extract them as isolated test case.

* github ci: added "Confluent Test"
* Upgrade confluent from 6.0 to 7.6
* Replace deprecated --zookeeper option to --bootstrap-server to
  create kafka topics


daipom (approved):
LGTM. Thanks!

PR #645, opened 2024-06-18
msi: update bundled ruby to 3.2.4

kenhys:
ref. https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-2-4-released/

drop already merged patch.

See https://github.com/oneclick/rubyinstaller2/pull/375


daipom (approved):
LGTM. Thanks.
We need to fix CGroupV1 tests...

kenhys:
> We need to fix CGroupV1 tests...

It seems that we need to give lxd up.

kenhys:
Let's switch to incus in another PR.

daipom:
I don't understand this cause...
In my local Ubuntu 20.04, I can install CGroup v1 images.
Has the specification (related to CGroup) of the runner Ubuntu 20.04 been changed?

kenhys:
As previously same workflow worked, so I guess that GitHub runner configuration such as remote server access or something  was changed or blocked, but few clue to identify it. (I don't have enough experience) :confused:

fluent-package/config.rb
```diff
@@ -23,20 +23,19 @@
...
 # https://rubyinstaller.org/downloads/ (7-ZIP ARCHIVES)
-BUNDLED_RUBY_INSTALLER_X64_VERSION = "3.2.3-1"
-BUNDLED_RUBY_INSTALLER_X64_SHA256SUM = "cf106d718499125a55dff82daee40e7744ed224694c96036765b84a3ac4f654c"
+BUNDLED_RUBY_INSTALLER_X64_VERSION = "3.2.4-1"
+BUNDLED_RUBY_INSTALLER_X64_SHA256SUM = "20f4c87fea871cadb3286da6c46cd018c12b9d444f0c40b3332b3f2939ce31b9"
 
 # Files under rubyinstaller/ are patches for RubyInstaller's binary package.
 # Other patches for Ruby's source tree which can be shared with BUNDLED_RUBY_PATCHES.
 BUNDLED_RUBY_INSTALLER_PATCHES = [
-  ["rubyinstaller/0001-Avoid-crash-on-invalid-registry-key-while-detecting-.patch", [">= 0"]],
```
ashie, line 41: Could you remove also the patch file itself?

kenhys: Thanks, I've removed it.

daipom:
> As previously same workflow worked, so I guess that GitHub runner configuration such as remote server access or something was changed or blocked, but few clue to identify it. (I don't have enough experience) 😕

Thanks. I see.
Anyway, it looks like we should try Incus.

kenhys:
rebased with master. expected CI will pass.

daipom:
>  msi: update bundled ruby to 3.2.4

Isn't this for the entire Ruby version in the package?

daipom (approved):
Thanks! LGTM.

PR #646, opened 2024-06-19
github: update remote image server

kenhys:
Since the Linux Containers project has made the decision to phase out access to linuxcontainers.org image server for LXD users, so we need to switch to new image server to use cgroup v1 images.

See
* https://discuss.linuxcontainers.org/t/important-notice-for-lxd-users-image-server/18479
* https://discourse.ubuntu.com/t/new-lxd-image-server-available-images-lxd-canonical-com/43824

### DIFF
--- a/.github/workflows/yum.yml
+++ b/.github/workflows/yum.yml
@@ -130,6 +130,10 @@ jobs:
         with:
           name: packages-${{ matrix.rake-job }}
       - uses: canonical/setup-lxd@v0.1.1
+      - name: Update image server
+        run: |
+          lxc remote remove images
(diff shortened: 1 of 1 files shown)


kenhys:
It may be enough to update remote image server for fixing CI failure, so I'll check it in this PR before switching to incus.

daipom:
Thanks!! Great!!

kenhys:
In out use case, it seems that it is enough just to switch remote image server to https://images.lxd.canonical.com.

kenhys:
Marked as resolved
https://github.com/orgs/community/discussions/126676

daipom (approved):
LGTM.

PR #649, opened 2024-06-28
github: restrict execution timeout

kenhys:
There is no need to waste CI resources.

ref.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes


daipom (approved):
LGTM

PR #650, opened 2024-06-28
fluent-package v5.0.4

kenhys:
As LTS update, basically updates patch versions.

This is a maintenance release to patch version of Ruby.


kenhys:
#617 should be merged before it.

daipom (approved):
LGTM. Thanks!

PR #651, opened 2024-07-01
rpm: drop centos:7 to collect artifacts

kenhys:
CentOS:7 amd64 binary can be built but not for arm64 anymore. It seems that arm64 specific hotfix is required.
(but such a effort is not maintainable anymore because of EOL)


kenhys:
I've reconsidered that it's time to drop amd64 too.

daipom (approved):
LGTM. Thanks!

PR #653, opened 2024-07-01
Improve test script

kenhys:
* set noninteractive not to block
* do not set UID explicitly on ubuntu
* switch to baseurl (only for testing with local container)

### DIFF
--- a/fluent-package/test-install-in-docker.sh
+++ b/fluent-package/test-install-in-docker.sh
@@ -82,6 +82,7 @@ DNF=dnf
 ID=$(cat /etc/os-release | grep "^ID=" | cut -d'=' -f2)
 case $ID in
     debian|ubuntu)
+	export DEBIAN_FRONTEND=noninteractive
 	CODENAME=$(cat /etc/os-release | grep VERSION_CODENAME | cut -d'=' -f2)
 	case $CODENAME in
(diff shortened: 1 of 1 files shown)


kenhys:
#654 should be merged in advance.

daipom:
I think we need #655 to make this test succeed.

kenhys:
As #654 was merged, rebased it.

fluent-package/test-install-in-docker.sh
```diff
@@ -14,10 +14,11 @@ set -e
 
 function setup_apt_user()
 {
+    export DEBIAN_FRONTEND=noninteractive
```
daipom, line 17: Is this for safety?
Should we do this around the following code instead of in the function if it is not limited to user setup?

https://github.com/fluent/fluent-package-builder/blob/7991bbb68ccf36439e200b208b2d2d8cb02d7b63/fluent-package/test-install-in-docker.sh#L92-L93

daipom (approved):
I have commented on a minor point, but it looks good to me!

fluent-package/test-install-in-docker.sh
kenhys: Surely it may be better to do so.
Thanks!

PR #654, opened 2024-07-02
github: drop CentOS 7 from GitHub workflow

kenhys:
CentOS 7 has reached EOL, so drop support for building it.

NOTE: keep build recipe for a while


daipom (approved):
LGTM. Thanks!

PR #655, opened 2024-07-03
test: fix experimental install test

daipom:
`TARGET` should not be used anymore.

The current test always installs the current latest version.
After that, the test updates it to the experimental.
So, the `TARGET` is meaningless, and the check always fails.

### DIFF
--- a/fluent-package/test-install-in-docker.sh
+++ b/fluent-package/test-install-in-docker.sh
@@ -59,23 +59,8 @@ exit 0
     su - $USER
 }
 
-function check_installed_version()
-{
-    VERSION=$1
-    case $VERSION in
-	*$TARGET*)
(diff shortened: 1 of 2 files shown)


kenhys:
test-verify-repo.sh is intended to be used to check whether repository metadata is correctly updated or not.
(before deploying experimental to LTS)

To verify it, firstly current version is installed, then upgrade to experimental's latest version.
If it can not be done correctly, we should not deploy them.

> So, the TARGET is meaningless, and the check always fails.

No. it is required to ensure whether the specified version is installable or not.

daipom:
I see! Sorry for confusing you.
So, `Target` is the latest version in the experimental repo and means the expected version after the update.

```
./test-verify-repo.sh 5.0.4
```

PR #656, opened 2024-07-05
Fix a bug that excessively block to launch

kenhys:
In the previous versions, guard duplicated instance mechanism was introduced.

But it should not block -h/--help and so on because user may refer command line option in such a case.


daipom (approved):
LGTM.
Thanks!

PR #657, opened 2024-07-05
msi: update Nokogiri to latest version

kenhys:
* Fixed CVE-2024-34459 in v1.16.5

Actually Nokogiri is not affected, but security scannaers may claim it.


fluent-package/Gemfile.lock
daipom, line 1: Should not we update here as well?

https://github.com/fluent/fluent-package-builder/blob/00e4771e06f596ef72842f6a72afd3662334e594/fluent-package/Gemfile.lock#L199-L201

kenhys: Thanks,

daipom (approved):
LGTM. Thanks!

PR #665, opened 2024-07-25
deb rpm: group logging

kenhys:
To investigate build log effectively, grouping log is better to do.


kenhys:
Verbose logs are grouped like this.

image

daipom (approved):
LGTM! Thanks!

PR #668, opened 2024-08-05
Drop centos:7 from default test target

kenhys:


daipom (approved):
LGTM. Thanks!

kenhys:
piuparts failure is unexpected, but it has nothing to do with this PR.
Let's fix in another PR.

PR #669, opened 2024-08-05
Allow to test more targets 5.x and 5.x (LTS)

kenhys:
Not only test/experimental, test standard version and LTS too.


daipom (approved):
LGTM. Thanks!

kenhys:
piuparts failure is known, and out-of-scope in this PR.

PR #670, opened 2024-08-05
rpm: unify to signing key

kenhys:
By PR#580, signing key was changed by default, but it doesn't follow for rpm package signing yet.


daipom (approved):
LGTM. Thanks!

kenhys:
piuparts failure is out-of-scope in this PR.

PR #672, opened 2024-08-06
Update supported platform example in README.md

kenhys:
As we won't update centos:7 package and supported ubuntu 24.04 recently.


daipom (approved):
LGTM. Thanks!

PR #674, opened 2024-08-21
Rakefile: install ruby gems with parallel

Watson1978:
Seems It appears to be taking a long time to install ruby gems on windows environment. So this patch will introduce parallel install to reduce installation time.

### DIFF
--- a/fluent-package/Rakefile
+++ b/fluent-package/Rakefile
@@ -436,7 +436,7 @@ class BuildTask
         gem_home = ENV["GEM_HOME"]
         ENV["GEM_HOME"] = gem_staging_dir
         ENV["INSTALL_GEM_FROM_LOCAL_REPO"] = "yes"
-        sh(bundle_command, "_#{BUNDLER_VERSION}_", "install")
+        sh(bundle_command, "_#{BUNDLER_VERSION}_", "install", "-j4")
         # Ensure to install binstubs under /opt/td-agent/bin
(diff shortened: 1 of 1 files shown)


daipom:
Thanks! Great job!

daipom:
#675 will solve the failing CI.

fluent-package/Rakefile
```diff
@@ -436,7 +436,7 @@ class BuildTask
         gem_home = ENV["GEM_HOME"]
         ENV["GEM_HOME"] = gem_staging_dir
         ENV["INSTALL_GEM_FROM_LOCAL_REPO"] = "yes"
-        sh(bundle_command, "_#{BUNDLER_VERSION}_", "install")
+        sh(bundle_command, "_#{BUNDLER_VERSION}_", "install", "-j4")
```
kenhys, line 439: Good catch!

But it is not suitable to directly specify "-j4" (fixed value) in this context.
How about using `Etc.nprocessors`?

kenhys (changes requested):

fluent-package/Rakefile
Watson1978: I changed to use `Etc.nprocessors` at https://github.com/fluent/fluent-package-builder/pull/674/commits/c304360364410882f169a70d2e0b547d92d1de73
Thanks!!

kenhys (approved):
LGTM.

kenhys:
Thanks!

PR #678, opened 2024-09-19
deb rpm: add helper script to generate index.html for hosting site

kenhys:
Currently, we use https://packages.treasuredata.com hosted on AWS
which is powered by Treasure Data Inc.

Recently, we coordinated with CNCF which provides access to Cloudflare
R2 resource. As no S3Browser functionality, so need to replace it with
alternative index.html.

### DIFF
--- a/fluent-package/make-index-html.rb
+++ b/fluent-package/make-index-html.rb
@@ -5,12 +5,14 @@ require 'optparse'
 
 options = {
   site: "https://fluentd.cdn.cncf.io",
-  verbose: false
+  verbose: false,
+  channel: %w(5 lts test)
 }
 
 opt = OptionParser.new
 opt.on("-s", "--site URL") { |v| options[:site] = v }
 opt.on("-v", "--verbose") { options[:verbose] = true }
+opt.on("-c", "--channel TARGET_CHANNEL") { |v| options[:channel] = v.split(',') }
 top_dir = opt.parse!(ARGV).first
 
 unless File.exist?(top_dir)
@@ -20,7 +22,7 @@ end
 
 template_path = File.expand_path("#{File.basename(__FILE__, ".rb")}.erb")
 puts "Template path: #{template_path}"
-%w(5 lts test).each do |channel|
+options[:channel].each do |channel|
   search_path = Pathname.new("#{top_dir}/#{channel}")
   Find.find(search_path).each do |path|
     Find.prune unless FileTest.directory?(path)
--- a/fluent-package/make-index-html.rb
+++ b/fluent-package/make-index-html.rb
@@ -4,15 +4,13 @@ require 'pathname'
 require 'optparse'
 
 options = {
-  :site => "https://fluentd.cdn.cncf.io",
-  :verbose => false
+  site: "https://fluentd.cdn.cncf.io",
+  verbose: false
 }
 
 opt = OptionParser.new
-opt.on("-s", "--site [URL]") { |v| options[:site] = v }
-opt.on("-v", "--verbose") { |v|
-  options[:verbose] = v
-}
+opt.on("-s", "--site URL") { |v| options[:site] = v }
+opt.on("-v", "--verbose") { options[:verbose] = true }
 top_dir = opt.parse!(ARGV).first
 
 unless File.exist?(top_dir)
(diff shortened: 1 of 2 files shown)


kenhys:
Preview of index.html

image

fluent-package/make-index-html.rb
```diff
@@ -0,0 +1,49 @@
...
+template_path = File.expand_path("#{File.basename(__FILE__, ".rb")}.erb")
+puts "Template path: #{template_path}"
+%w(5 lts test).each do |channel|
+  search_path = Pathname.new("#{top_dir}/#{channel}")
+  Find.find(search_path).each do |path|
+    Find.prune if path.end_with?(".deb")
+    Find.prune if path.end_with?(".msi")
+    Find.prune if path.end_with?(".gz")
+    Find.prune if path.end_with?(".bz2")
+    Find.prune if path.end_with?(".rpm")
```
ashie, line 32: Is there any reason to check suffix?
It seems that we can skip all regular files:
```suggestion
    Find.prune unless FileTest.directory?(path)
```

fluent-package/make-index-html.rb
```diff
@@ -0,0 +1,49 @@
...
+top_dir = opt.parse!(ARGV).first
+
+unless File.exist?(top_dir)
+  puts "#{top_dir} not found"
+  exit 1
+end
+
+template_path = File.expand_path("#{File.basename(__FILE__, ".rb")}.erb")
+puts "Template path: #{template_path}"
+%w(5 lts test).each do |channel|
```
ashie, line 25: How about past td-agent releases?

ashie: > How about past td-agent releases?

Ah, it's not needed after we generate them once since they are already EOL.

fluent-package/make-index-html.rb
kenhys: Thanks!

ashie (approved):
:+1:

kenhys:
I'll wait feedback from @daipom

fluent-package/make-index-html.rb
```diff
@@ -0,0 +1,45 @@
+require 'erb'
+require 'find'
+require 'pathname'
+require 'optparse'
+
+options = {
+  :site => "https://fluentd.cdn.cncf.io",
+  :verbose => false
+}
```
daipom, line 9:
```suggestion
options = {
  site: "https://fluentd.cdn.cncf.io",
  verbose: false
}
```

fluent-package/make-index-html.rb
```diff
@@ -0,0 +1,45 @@
...
+
+opt = OptionParser.new
+opt.on("-s", "--site [URL]") { |v| options[:site] = v }
```
daipom, line 12:
```suggestion
opt.on("-s", "--site URL") { |v| options[:site] = v }
```

`[]` makes the value optional. It means the value can be `nil`.
It appears that we should not allow `nil` here.

fluent-package/make-index-html.rb
```diff
@@ -0,0 +1,45 @@
...
+opt.on("-v", "--verbose") { |v|
+  options[:verbose] = v
+}
```
daipom, line 15:
```suggestion
opt.on("-v", "--verbose") { options[:verbose] = true }
```

It can be simpler like this.

fluent-package/make-index-html.rb
```diff
@@ -0,0 +1,45 @@
...
+
+template_path = File.expand_path("#{File.basename(__FILE__, ".rb")}.erb")
+puts "Template path: #{template_path}"
+%w(5 lts test).each do |channel|
+  search_path = Pathname.new("#{top_dir}/#{channel}")
+  Find.find(search_path).each do |path|
+    Find.prune unless FileTest.directory?(path)
+    Find.prune if path.end_with?("repodata")
+    Find.prune if path.end_with?("dists")
+    if %w(windows x86_64 aarch64 fluent-package).any? { |dir| path.end_with?(dir) }
```
daipom, line 31:
```suggestion
Dir.glob("#{top_dir}/{5,lts,test}/**/{windows,x86_64,aarch64,fluent-package}/") do |dir|
```

Do we need `Find`?
How about simply using `glob` like this?

fluent-package/make-index-html.rb
```diff
@@ -0,0 +1,45 @@
...
+      puts "Updating: #{path} ..." if options[:verbose]
+      index_path = File.expand_path(File.join(path, "index.html"))
+      Dir.chdir(path) do
+        files = Dir.glob(["*.deb", "*.msi", "*.rpm"])
+        relative_path = path.sub(top_dir, "")
```
daipom, line 36:
```suggestion
        relative_path = Pathname.new(path).relative_path_from(top_dir).to_s
```

This would be safer.

daipom (approved):
I have checked mainly the Ruby scripts.
It could be simpler, as I commented on below, but I don't think it would be a problem as it is.

fluent-package/make-index-html.rb
kenhys: Thanks, this approach may be better.

PR #681, opened 2024-10-02
ci: add update tests about service status

daipom:
There is specification difference between DEB and RPM.

At least, there should be tests to clarify the current specification.

The current specification:

* DEB
  * The service is **enabled** by default.
  * The service **starts automatically after new install**.
  * The service starts automatically **when it was active OR enabled before update**.
* RPM
  * The service is **disabled** by default.
  * The service **does not start automatically after new install**.
  * The service starts automatically **only when it was active before update**.

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -117,7 +117,7 @@ jobs:
           ${{ matrix.test-docker-image }} \
           /fluentd/fluent-package/apt/binstubs-test.sh
   test:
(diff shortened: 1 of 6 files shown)


daipom:
Rebased to the latest master.

daipom:
I have refactored the codes.
Confirming CI...

Watson1978 (approved):
👍🏻

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #682, opened 2024-10-03
Embed fluent package info into environment variable

kenhys:
Embed fluent package info into environment variable

It is aimed to detect current running process's fluentd version and
so on from /proc/id/environ.

This fix introduce the following environment variables which contains
similar information as same as /usr/sbin/fluentd -v

* FLUENT_PACKAGE_VERSION
* FLUENT_PACKAGE_FLUENTD_VERSION
* FLUENT_PACKAGE_FLUENTD_REVISION

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -133,6 +133,7 @@ jobs:
         test-file:
           - "update-from-v4.sh"
           - "update-to-next-version-with-backward-compat-for-v4.sh"
+          - "update-to-next-version.sh"
           - "downgrade-to-v4.sh"
           - "install-newly.sh local"
           - "install-newly.sh v5"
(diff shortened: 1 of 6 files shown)


daipom:
Thanks!
We need some tests for this feature.
Also, I'm concerned about update.
When updating to the next version, does this env var be updated?
It would be better to check it by CI.

kenhys:
TODO: check whether env will be updated correctly. it should care that case.

kenhys:
added test case to check environment variable.

kenhys:
unexpectedly, update-to-next-version.sh was not enabled yet.

kenhys:
It seems the regression of https://github.com/fluent/fluent-package-builder/pull/636

kenhys:
deb, rpm tests are now all green.

kenhys:
As #684 was merged, rebased it.

Watson1978 (approved):
LGTM :)

PR #685, opened 2024-10-04
Fix deprecated usage of ERB.new

kenhys:
It fixes the following warning (Since Ruby 3.1):

  Rakefile:168: warning: Passing safe_level with the 2nd argument of ERB.new is deprecated. Do not use it, and specify other arguments as keyword arguments.
  Rakefile:168: warning: Passing trim_mode with the 3rd argument of ERB.new is deprecated. Use keyword argument like ERB.new(str, trim_mode: ...) instead.

This does not change the content of fluent-package.


daipom (approved):
LGTM. Thanks!

PR #686, opened 2024-10-07
Add logic to switch auto/manual service restart during upgrading

kenhys:
auto: Automatically restart service without downtime.

manual: Manually restart service by user. If user select manual,
  suppress auto-restart with automatically generated service hook
  %post and %systemd_postun_with_restart.


kenhys:
should be working on feature-nodowntime branch, so I'll close this PR.

PR #691, opened 2024-10-18
deb: fix inconsistent phase comparison

kenhys:
fluentd_auto_restart will be launched during
"configure" phase, so there is no need to check that action.

This is occurred by migrating restart logic from postrm to postinst.


Watson1978 (approved):
👍🏻

PR #692, opened 2024-10-22
deb rpm: add helper script to repack debug packages

kenhys:
It is useful to investigate maintainer scripts.

### DIFF
--- a/fluent-package/debug-package-converter.sh
+++ b/fluent-package/debug-package-converter.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/bash
+
+#
+# Helper script to generate package which
+# can enable debug verbose message in maintainer scripts.
+# It helps to trace internal behavior install/upgrade/remove
+# and so on.
+#
+# Usage: debug-package-converter.sh [debug] PATH_TO_DEB_OR_RPM
+#
+# e.g.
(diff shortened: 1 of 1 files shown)


fluent-package/test-repack-package.sh
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/bash
+
+#
+# Usage: test-repack-package.sh [OPTIONS] PATH_TO_DEB_OR_RPM
+#
+# e.g.
+# $ test-repack-package.sh debug fluent-package-6.0.0-1.el8.x86_64.rpm
+# $ test-repack-package.sh debug fluent-package_6.0.0-1_amd64.deb
+#
```
daipom, line 9: Thanks for this!
It looks convenient, but I don't understand what we can do by using this script...
Could you please reinforce these usage comments?

kenhys: Thanks, I've updated explanation

daipom:
Thanks! I see!
This script adds debug messages to the existing package!

daipom:
Should the script name be `test-repack-package`?
To me, it looks like a test script, not a helper script.
Wouldn't a name like `repack-package-with-debug-message`, for example, be easier to understand?
(This might be long, though...)

daipom (approved):
LGTM. Thanks!

PR #693, opened 2024-10-23
deb: simplify helper script

kenhys:
Added if guard to check whether specified hook script exist or not.


daipom (approved):
LGTM. Thanks!

PR #694, opened 2024-10-23
deb rpm: fix v4 migration with restarting service

kenhys:


kenhys:
It might be lxd is a bit freaky?

kenhys:
Changes against apt/systemd-test/*.sh is a workaround until /etc/needrestart/conf.d/ fix will be merged into.

daipom:
Shouldn't we fix this on the master branch?

kenhys:
TODO: re-target master branch.

PR #696, opened 2024-10-29
rpm ci: collect diagnostic message for cgroup v1

kenhys:
lxc launch will fail (/var/snap/lxd/common/lxd/unix.socket: connect: permission denied) in some cases, but not sure why yet.

To investigate further more, try to collect snap information when workflow was failed.


daipom (approved):
Thanks!

PR #698, opened 2024-10-29
ci: add cancel-in-progress setting

kenhys:
see https://github.com/fluent/fluentd/pull/4689

ref. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs


daipom (approved):
LGTM. Thanks!

PR #699, opened 2024-10-30
ci: add test to update without data lost

Watson1978:
This PR adds testing to ensure that package upgrades succeed without data loss.
At present, the following plugins will be checked:

- in_tcp
- in_udp
- in_syslog


daipom (approved):
LGTM! Thanks!!

PR #703, opened 2024-10-30
ci: test deb migration from repository

daipom:
Behavior may be different from installing from a deb file.
(Some tests should fail.)


kenhys (approved):

kenhys:
TODO: we need to fix the issue via installing local and remote.

PR #704, opened 2024-10-30
ci: cache package to reduce build time

kenhys:


kenhys:
rpm cached:
https://github.com/fluent/fluent-package-builder/actions/runs/11586846546/job/32259162192

deb cached:
https://github.com/fluent/fluent-package-builder/actions/runs/11586846546/job/32259162421

daipom:
Thanks! 
I think we should add this to the master branch because this is not directly related to this feature branch.

.github/workflows/apt.yml
```diff
@@ -46,7 +46,17 @@ jobs:
...
           sudo gem uninstall fileutils
+      - name: cache deb
+        uses: actions/cache@v4
+        id: cache-deb
+        with:
+          path: |
+            fluent-package/apt/repositories
+            fluent-apt-source/apt/repositories
+            fluent-lts-apt-source/apt/repositories
+          key: ${{ runner.os }}-cache-${{ matrix.rake-job }}-${{ hashFiles('**/config.rb', '**/Rakefile', '**/Gemfile*', '**/deb/*') }}
```
daipom, line 57: Using `hashFiles` is an excellent idea.
but is this enough? Don't we need to include more files?

* `fluent-package/templates/**`
* `fluent-package/debian/**`
* files under `fluent-package/apt/` except for test files, such as Dockerfile

kenhys: Thanks, take into account more files .

kenhys:
> I think we should add this to the master branch because this is not directly related to this feature branch.

Basically so, but it is effective to apply to this feature branch.

kenhys:
re-target to master.

PR #705, opened 2024-10-30
ci: show more debug message to track

kenhys:


daipom (approved):
Thanks for investigating this issue.

PR #708, opened 2024-10-31
fluent-package: improve repack helper script

kenhys:
* enable debug by default
* support any package and architecture
* untabify


daipom (approved):
Thanks!

PR #709, opened 2024-10-31
ci: install transitional package

kenhys:
Usually, transitional package (td-agent) will be installed when upgrading from v4. so it should be installed at the same time　
to test in practical use case.


daipom (approved):
Thansk!
Note: It does not change the test behavior, but we should install the transitional package for the tests.

PR #710, opened 2024-10-31
ci: Noble: enable install-newly test

daipom:
The package for Noble has already been released, so we should enable the tests.


kenhys (approved):
LGTM.

PR #712, opened 2024-11-01
ci: add test to update with auto / manual feature

Watson1978:
add test to update with auto / manual feature.


daipom (approved):
LGTM. Thanks!!

PR #715, opened 2024-11-07
ci: update the tests for no data lost

Watson1978:
Update the test to ensure :

* There is no issue even if an upgrade is performed that includes a Ruby major version update.
* There is no data loss even if data is sent until the main process is replaced.


daipom (approved):
Thanks!

PR #716, opened 2024-11-07
Backport: apt: fix broken piuparts test case (LTS)

kenhys:
Since v5.1.0, signing key was changed, so it should be included in archive key for piuparts.

In LTS branch, even though old signing key is still used, but it might better to migrate.

ref. https://github.com/fluent/fluent-package-builder/pull/671


daipom (approved):
LGTM. Thanks!

PR #717, opened 2024-11-07
github: fix target branch name

kenhys:
At first, assumed 3 branchs master, v5, lts-v5, but it might be enough for master (v5.x) and lts-v5 for a while.
Already branch name was changed, so follow it.


daipom (approved):
LGTM! Thanks!

PR #684, opened 2024-10-04
github: try to use more customized test labels

kenhys:
Currently, it is bit hard to distinct each Test jobs
on fluent-package-builder/actions/runs page.

This is because Test (CGroup V1) matrix.label matrix.test-file
is shown by default.

This commit try to make it more visibility on that page.

PR #720, opened 2024-11-07
Backport: github: try to use more customized labels (LTS)

kenhys:
Currently, it is bit hard to distinct each Test jobs on fluent-package-builder/actions/runs page.

This is because Test (CGroup V1) matrix.label matrix.test-file is shown by default.

This commit try to make it more visibility on that page.

ref. https://github.com/fluent/fluent-package-builder/pull/684


daipom (approved):
LGTM. Thanks!

PR #675, opened 2024-08-21
ci: fix failing Test (CGroup V1)

daipom:
It appears that now the default `remote images` is https://images.lxd.canonical.com.

I don't know why, but we can't remove `images`.
The following command does nothing, so this step fails.

    lxc remote remove images

PR #723, opened 2024-11-08
Backport: ci: fix failing Test (CGroup V1) (LTS)

kenhys:
It appears that now the default `remote images` is https://images.lxd.canonical.com.

I don't know why, but we can't remove `images`.
The following command do nothing, so this step fails.

    lxc remote remove images

ref. https://github.com/fluent/fluent-package-builder/pull/675


daipom (approved):
LGTM. Thanks!

PR #728, opened 2024-11-08
Backport: ci msi: cache package to reduce build time

kenhys:
Even though GitHub Actions' cache over limitations easily, it might better to cache only Windows.


Watson1978 (approved):
👍🏻

PR #729, opened 2024-11-08
Backport:  msi: set GEM_HOME / GEM_PATH in batch file

kenhys:
If users set custom GEM_HOME / GEM_PATH environment variables, the batch file will not work as expected. Therefore, this patch will set the appropriate values in the batch file.


Watson1978 (approved):
👍🏻

PR #731, opened 2024-11-11
Backport: ci: install transitional package

kenhys:
Usually, transitional package (td-agent) will be installed when upgrading from v4. so it should be installed at the same time　 to test in practical use case.

ref. https://github.com/fluent/fluent-package-builder/pull/709
 https://github.com/fluent/fluent-package-builder/pull/711


daipom (approved):
LGTM. Thanks!

PR #732, opened 2024-11-11
Backport ci deb: tweak downgrade test case

kenhys:
After removing fluent-package and td-agent,
fluentd.service file is linked to nonexistent file (masked), but td-agent service is still active unexpectedly, so explicitly stop that service.


daipom:
Thanks! Should we merge this into the master, and backport it to the v5 branch?

daipom (approved):
LGTM. Thanks!

PR #734, opened 2024-11-12
ci: add tests for downgrading without data lost

Watson1978:
This PR will add tests for downgrading without data lost.


kenhys (approved):
LGTM.

PR #735, opened 2024-11-14
package: improve verify helper

kenhys:
* show targets on startup
* show ID (e.g. debian:bookworm) when failed to compare versions
* skip test when "dummy" was specified for DEB_TARGETS/RPM_TARGETS
* logging for each test case (e.g. install-debian-bookworm-on-exp-lts-5.log)

### DIFF
--- a/fluent-package/test-verify-repo.sh
+++ b/fluent-package/test-verify-repo.sh
@@ -9,10 +9,14 @@
 
 function test_deb() {
     for d in $DEB_TARGETS; do
+	if [ $d = "dummy" ]; then
+	    continue
+	fi
 	for r in $REPO_TARGETS; do
 	    echo "TEST: on $d $r"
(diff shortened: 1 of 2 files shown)


fluent-package/test-verify-repo.sh
```diff
@@ -52,7 +60,12 @@ if [ -z "$REPO_TARGETS" ]; then
     REPO_TARGETS="exp/5 exp/lts/5"
 fi
 
-RESULTS=""
+echo "DEB_TARGETS: $DEB_TARGETS"
+echo "RPM_TARGETS: $RPM_TARGETS"
+echo "REPO_TARGETS: $REPO_TARGETS"
+sleep 3
+echo RESULTS=""
```
daipom, line 67: > sleep 3
> echo RESULTS=""

Could you tell me the purpose of these two lines?

kenhys: unexpectedly garbage was injected.

daipom (approved):
LGTM. Thanks!

PR #736, opened 2024-11-14
Backport: package: improve verify helper

kenhys:
* show targets on startup
* show ID (e.g. debian:bookworm) when failed to compare versions
* skip test when "dummy" was specified for DEB_TARGETS/RPM_TARGETS
* logging for each test case


daipom (approved):
Thanks!

PR #737, opened 2024-11-14
ci yum: tweak cgroup v1 failure

kenhys:
Before lxc launch container image, it might better to wait some extent for stabilizing CI as known tips.

It works well if you collect diagnostic message for it.

NOTE: this fix was already landed in feature-nodowntime branch.


daipom (approved):
Thanks!

PR #741, opened 2024-11-25
update Fluentd for 1.18 test

daipom:
Follow update of:

* https://github.com/fluent/fluentd/tree/restart-without-downtime-test-for-1.18


Watson1978 (approved):
👍🏻

PR #747, opened 2024-11-29
update Fluentd and ServerEngine and other dependencies

daipom:
* Update Fluentd to v1.18.0 (https://github.com/fluent/fluentd/releases/tag/v1.18.0).
* Update ServerEngine to v2.4.0 (https://github.com/treasure-data/serverengine/commit/c93a592bb9642cdb8880fe9366bbee37e5b764cd).
* Update the lock file by `bundle update` on Windows.

Note: Normally, we update dependencies at the time of release, but this time, we should do it first since we will add a new feature #713 that depends on these new versions.


Watson1978 (approved):
👍🏻

kenhys (approved):
LGTM.

PR #752, opened 2024-12-04
Remove unnecessary GEM_HOME/GEM_PATH settings

daipom:
Remove environment variables that prevent zero-downtime-restart feature of Fluentd v1.18.0. This is because an upgrade with a Ruby version update will not work correctly due to environment variables for the old Ruby.

Note: Cherry-picked 9762e1f4938a1448eef6f40bc69c05f1156b7c18 on `feature-nodowntime` branch.


Watson1978 (approved):
👍🏻

PR #755, opened 2024-12-05
ci: add test to check the service status with updating from v4

Watson1978:
This PR will add test to confirm that fluentd should be restarted with updating from v4

### DIFF
--- a/fluent-package/apt/systemd-test/update-from-v4.sh
+++ b/fluent-package/apt/systemd-test/update-from-v4.sh
@@ -9,6 +9,7 @@ sudo apt install -y curl ca-certificates
 curl -fsSL https://toolbelt.treasuredata.com/sh/install-${distribution}-${code_name}-td-agent4.sh | sh
 
 systemctl status --no-pager td-agent
+main_pid=$(eval $(systemctl show td-agent --property=MainPID) && echo $MainPID)
 
 # Generate garbage files
 touch /etc/td-agent/a\ b\ c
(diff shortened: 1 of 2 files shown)


daipom:
Looks like restart does not occur unexpectedly when updating v4 to v5.
(td-agent v4 (active) -> fluent-package v5)

fluent-package/apt/systemd-test/update-from-v4.sh
```diff
@@ -49,6 +50,9 @@ esac
 systemctl status --no-pager fluentd
 systemctl status --no-pager td-agent
 
+# Fluentd should be restarted when update from v4.
+test $main_pid -ne $(eval $(systemctl show fluentd --property=MainPID) && echo $MainPID)
```
daipom, line 54: Thanks for this test enhancement.

For some reason, the restart doesn't happen in deb.

Only when updating from v4 to v5, the behavior seems different.
(Normally, in deb, restart happens when the service was active before the update.)

I don't think it's as expected, but updating from v4 to v5 would normally be done after stopping the service, so this behavior would not be a significant issue.

In the meantime, please reverse this expectation and leave a comment.

fluent-package/apt/systemd-test/update-from-v4.sh
```diff
@@ -18,6 +19,9 @@ for d in $(seq 1 10); do
     touch /var/log/td-agent/$d.log
 done
 
+# When it updates from v4, the service needs to be stopped to restart fluentd after installation.
+sudo systemctl stop td-agent
```
daipom, line 23: I see! Looks good.
Could you please leave a comment that this behavior is unexpected for us, such as the following?

```suggestion
# When it updates from v4, the service needs to be stopped to restart fluentd after installation.
# Note: This behavior is not expected, and maybe it should be fixed.
# The cause is unknown. For deb, usually, restart should happen when the service was active before the update.
sudo systemctl stop td-agent
```

daipom (approved):
LGTM. Thanks!

PR #758, opened 2024-12-09
deb rpm: remove manual feature of zero-downtime-restart from uninstall phase

Watson1978:
We were going to support two methods of downgrading with zero downtime.

1. Running v6.x => Install v5.x package by overwriting
2. Running v6.x => Replace `FLUENT_PACKAGE_SERVICE_RESTART` value to `manual` => uninstall v6.x =>  Install v5.x

The second method is to keep the Fluentd process running after uninstalling it.
We decided to remove `2.` method because it may cause confusion for users.


daipom:
Thanks! Looks good!
Let's wait CI results.

daipom (approved):

daipom:
Thanks!

PR #759, opened 2024-12-09
rpm: revert suppressing systemd_post macro

daipom:
We don't need to suppress this macro because this macro handles preset, not restart.

I have confirmed that this macro of the package for RHEL 9 is expanded as follows.

```bash
if [ $1 -eq 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then
    # Initial installation
    /usr/lib/systemd/systemd-update-helper install-system-units fluentd.service || :
fi
```


Watson1978 (approved):
👍🏻

daipom:
Thanks!

PR #762, opened 2024-12-10
deb: improve process timing for safety and simplicity

daipom:
Points

* FROM-package just leaves tmp files if it supports the features.
* TO-package trigger the features if there are those tmp files.
* Thus, we don't need to check the version.
* Make installing plugin and restarting the same condition.

Before

1. from-prerm(upgrade): Do nothing.
2. to-preinst(upgrade): Collect plugin-list. Confirm version.
3. Install TO-package
4. from-postrm(upgrade): Do nothing.
5. Uninstall FROM-package
6. to-postinst(configure): Install plugin and restart if need.

After

1. from-prerm(upgrade):
   * Check auto or not.
   * Leave plugin-list and pid if need.
2. to-preinst(upgrade): Set tmp files for TO-package.
3. Install TO-package
4. from-postrm(upgrade): Clean tmp files of FROM-package.
5. Uninstall FROM-package
6. to-postinst(configure): Install plugin and restart if need.

### DIFF
--- a/fluent-package/templates/package-scripts/fluent-package/deb/preinst
+++ b/fluent-package/templates/package-scripts/fluent-package/deb/preinst
@@ -10,40 +10,25 @@ set -e
 # for details, see https://www.debian.org/doc/debian-policy/ or
 # the debian-policy package.
 
-local_base_plugins=/tmp/<%= package_dir %>/.local_base_plugins
-migrate_local_plugins() {
-    if [ "$FLUENT_PACKAGE_SERVICE_RESTART" != auto ]; then
-        return
+mark_auto_restart_ready() {
+    # Copy tmp files made by FROM-side because they should be cleaned postrm of FROM-side.
+    # (To ensure cleanup, tmp files should be cleaned by its own-side).
+    # The sequence of these tmp files is as follows:
+    #   1. FROM-prerm(upgrade): Leave tmp files if need.
+    #   2. TO-preinst(upgrade): Copy tmp files for TO-side.
+    #   3. FROM-postrm(upgrade): Clean tmp files of FROM-side(1.).
+    #   4. TO-postinst(configure): Use and clean tmp files of TO-side(2.).
+    if [ -f "/tmp/<%= package_dir %>/.plugin_list" ]; then
+        cp "/tmp/<%= package_dir %>/.plugin_list" "/tmp/<%= package_dir %>/.previous_plugin_list"
     fi
-    if [ ! -d /run/systemd/system ]; then
-        # tmpfiles.d owns /tmp/<%= package_dir %>, but not created without systemd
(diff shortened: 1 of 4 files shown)


fluent-package/templates/package-scripts/fluent-package/deb/preinst
```diff
@@ -10,40 +10,19 @@ set -e
 # for details, see https://www.debian.org/doc/debian-policy/ or
 # the debian-policy package.
 
-local_base_plugins=/tmp/<%= package_dir %>/.local_base_plugins
-migrate_local_plugins() {
-    if [ "$FLUENT_PACKAGE_SERVICE_RESTART" != auto ]; then
-        return
+check_auto_restart() {
+    if [ -f "/tmp/<%= package_dir %>/.plugin_list" ]; then
+        cp "/tmp/<%= package_dir %>/.plugin_list" "/tmp/<%= package_dir %>/.previous_plugin_list"
```
kenhys, line 15: Could you put a proper explanation "why" copy happen here?
(it might be enough to copy it from commit message)

How about name `mark_auto_restart_ready()` because let it in postinst stage.

fluent-package/templates/package-scripts/fluent-package/deb/prerm
```diff
@@ -16,18 +16,52 @@ set -e
...
+    if [ $pid -eq 0 ]; then
+        echo "Do not use auto restart because the service is not active"
+        return
+    fi
+
+    . "/etc/default/<%= service_name %>"
+    echo "FLUENT_PACKAGE_SERVICE_RESTART: $FLUENT_PACKAGE_SERVICE_RESTART"
+    if [ "$FLUENT_PACKAGE_SERVICE_RESTART" != auto ]; then
+        return
+    fi
```
kenhys, line 45: Could you comment a possibility that there is a case that these temporary files are left without clean-up.

Watson1978 (approved):
👍🏻

fluent-package/templates/package-scripts/fluent-package/deb/prerm
kenhys: Opps, above comment is nonsense. sorry.

fluent-package/templates/package-scripts/fluent-package/deb/preinst
daipom: Thanks for your review!
I see!

Fixed: https://github.com/fluent/fluent-package-builder/compare/0b673de8ec94dc360ed5df215b1af8f1c3749568..be1f3fe0d4e568809aa938e46f3444a94fe0ca42

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #768, opened 2024-12-12
Enable parallel download artifacts

kenhys:
In the previous versions, it download artifacts from GitHub in serial not parallel, so it takes huge amount of times.

In this version, it tries to download them in parallel, so it fixes above issue.


daipom (approved):
LGTM! Thanks!

PR #771, opened 2024-12-26
ci: check whether needrestart was suppressed for v5

kenhys:
As v5.2.0 was released,  needrestart suppression should be also checked.

Additionally extract check logic as common helper function.


daipom (approved):
Thanks!

PR #772, opened 2025-01-08
Use armored keyring for APT repository

kenhys:
Since apt 2.9.16, it rejects the keybox format files. So we should use
    the armored format instead.

ref. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1088656#35

    > The only format which is GnuPG specific, are for example its internal
    > keybox keyrings, which I don't think has ever been used as any kind
    > of public interchange format.

In practical use case, Debian 13 (trixie) or Ubuntu 25.04 will be
    affected. Without it, it causes verification error with apt-get update.

### DIFF
--- a/fluent-apt-source/Rakefile
+++ b/fluent-apt-source/Rakefile
@@ -34,7 +34,7 @@ class FluentdAptSourcePackageTask < PackageTask
   end
 
   def repository_version
-    "2023.6.29"
+    "2025.1.8"
   end
 
   def repository_name
--- a/fluent-apt-source/Rakefile
+++ b/fluent-apt-source/Rakefile
@@ -86,11 +86,8 @@ class FluentdAptSourcePackageTask < PackageTask
 
   def apt_targets_default
     [
-      "debian-buster",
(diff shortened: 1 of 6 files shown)


kenhys:
Steps to verify:

*  apt install -y vim curl sudo
*  execute install script which is documented on fluent.org
*  change /etc/apt/sources.list.d/fluent.sources or fluent-lts.sources, point to test/experimental
*  apt update
* apt upgrade -y, fluent-apt-source or fluent-lts-apt-source should be updated.
* apt update should be succeed

Result:

* bullseye v5 OK
* bullseye lts OK
* bookworm v5 OK
* bookworm lts NG (fluent-package depends libssl1.1, it should be fixed as another issue)
* focal v5 OK
* focal lts OK
* jammy v5 OK
* jammy lts OK
* noble v5 OK
* noble lts OK

Watson1978 (approved):
👍🏻

daipom (approved):
LGTM. Thanks!

Issue #602, opened 2023-11-28
Keep fluentdwinsvc commandline option during upgrading


kenhys:
# Problem

When upgrading fluent-package on windows, fluentdwinsvc configuration is always reset to factory default.

# Expected

If user was configured fluentdwinsvc configuration, it is better to keep it.

daipom:
Note

* #601

Related to this, when we enable automatic service start, it is disabled again after updating.

Example(Fluent Package Command Prompt with administrative privilege):
```cmd
$ sc config fluentdwinsvc start=delayed-auto
```

daipom:
Want to backport to 5.0.6.

PR #776, opened 2025-01-29
windows: backup and restore registry with update install

Watson1978:


daipom:
Thanks!
Note: this fixes:

* #602

daipom:
Looks like this restores the all registry values in `HKLM\System\CurrentControlSet\Services\fluentdwinsvc`.
So, this restores `ImagePath` too?

```
ImagePath    REG_EXPAND_SZ    "C:/opt/fluent/bin/ruby.exe" -C "C:/opt/fluent/lib/ruby/gems/3.2.0/gems/fluentd-1.18.0/lib/fluent/command/.."  winsvc.rb --service-name fluentdwinsvc
```

If so, I have concerns about whether the service will work correctly after the update.
CI seems fine, I wonder why.

Watson1978:
> So, this restores ImagePath too?

Indeed. I think there is two solution for that.

1. Remove `"ImagePath"` section from backup registry file before storing.
2. Or, Remove `"ImagePath"` in `HKLM\SYSTEM\CurrentControlSet\Services\fluentdwinsvc` before backup. Aftter then, it back up `HKLM\SYSTEM\CurrentControlSet\Services\fluentdwinsvc` into fluentdwinsvc.reg and restore it.

PR #777, opened 2025-01-30
msi: keep some registry values with update

### LINKED ISSUE #602: Keep fluentdwinsvc commandline option during upgrading

daipom:
* Fixes #602 
* Another idea for #776

~~This adds a new `Feature` to keeps the following registry values (in `HKLM\System\CurrentControlSet\Services\fluentdwinsvc`) when updating:~~
This keeps the following registry values (in `HKLM\System\CurrentControlSet\Services\fluentdwinsvc`) when updating:

* `Start`
* `DelayedAutostart`
* `fluentdopt`

~~Users can disable this feature when installing.~~

~~Note: The integer(DWORD) values have `#` prefix, so it cannot be written as is. So this defines each value pattern.~~
(ref https://wixtoolset.org/docs/v3/xsd/wix/registrysearch/)
@Watson1978 found we can write such a raw value with a prefix by `string` Type. Thanks!


daipom:
2025-01-29_22h58_05

daipom:
Points:

* We might need to give more consideration to the Feature name: `Feature to keep some registry values when updating`.
* The advantages compared to #776 are as follows:
  * The operation is completed on the new package side.
  * No need to handle a temporary file.
* The disadvantages compared to #776 are as follows:
  * Need to create a code for each registry value.
  * Hard to handle the prefix. This makes the code complicated.

daipom:
Changed following:

* Dropped the idea to add a new Feature for this.
* Fix some indends.
* Use simpler solution. (Thanks @Watson1978 !)

fluent-package/msi/source.wxs
```diff
@@ -136,6 +136,7 @@
       <ComponentRef Id="FluentConf" />
       <ComponentRef Id="FluentAcl" />
       <ComponentRef Id="FluentdBat" />
+      <ComponentGroupRef Id="KeepSomeRegistryValues" />
```
kenhys, line 139:
```suggestion
      <ComponentGroupRef Id="KeepUserRegistryConfiguration" />
```

How about naming with "What" ?

Watson1978 (approved):

fluent-package/msi/source.wxs
```diff
@@ -159,6 +160,46 @@
...
+
+    <!-- Restore some registry values -->
+    <ComponentGroup Id="KeepSomeRegistryValues" Directory="FLUENTPROJECTLOCATION">
+      <!--
+        CAUTION: We cannot use `integer` type for values being read by RegistrySearch. Use `string` instead.
+        RegistrySearch reads a raw value, and it can have a prefix according to the data type.
+        For example, integer (DWORD) value has a `#` prefix.
+        It appears that RegistryValue can use `string` type to handle such raw values.
+        If using `integer` type, it appears that it is forced to be recognized as `REG_SZ` type after all, and it breaks the service.
+       -->
```
kenhys, line 182: Hmm...

kenhys (changes requested):
Basically LGTM.

fluent-package/msi/source.wxs
daipom: Thanks! Fixed!

fluent-package/msi/source.wxs
kenhys:
```
CAUTION: We cannot use `integer` type for values being read by RegistrySearch. Use `string` instead.
DETAIL: RegistrySearch reads a raw value and the value could have a prefix according to the data type.
For example, integer (DWORD) value has a `#` prefix.
If using `integer` type, it appears that it is forced to be recognized as `REG_SZ` type after all, and it make hard to handle from wixtoolset (.wxs).
(It should be converted as Value=N without '#' prefix in this case)
Thus, it is better to use `string` type to handle such raw values in this context because of no need to care about '#' prefix.
```

daipom: Thanks!
I tried to put them together!
a35a13fa52fe6976cfd2ea69d7df595e6df700fb

daipom:
This PR restores the registry values as expected, but somehow, `Auto (Delayed)` setting still becomes simple `Auto` when updating.
There may be another setting than `DelayedAutostart`.

For now, let this issue be outside the scope of this PR.

kenhys (approved):
LGTM.

PR #779, opened 2025-02-03
Backport(v5)  msi: keep some registry values with update (#777)

### LINKED ISSUE #602: Keep fluentdwinsvc commandline option during upgrading

Watson1978:
This is backport #777 to v5 branch.

----

* Fixes #602
* Another idea for #776

This keeps the following registry values (in
`HKLM\System\CurrentControlSet\Services\fluentdwinsvc`) when updating:

* `Start`
* `DelayedAutostart`
* `fluentdopt`

(ref https://wixtoolset.org/docs/v3/xsd/wix/registrysearch/) @Watson1978 found we can write such a raw value with a prefix by `string` Type. Thanks!


daipom:
Thanks! Waiting CI.

daipom (approved):
LGTM. Thanks!

Note: The CI failure for Amazon Linux 2023 has nothing to do with this PR.
We will handle it in another PR.

PR #784, opened 2025-02-03
amazonlinux-2023: ensure to install `which` command

Watson1978:
* It appears that the recent Amazon Linux 2023 image stops providing `which`.
* We need `which` to build `librdkafka 1.9.0` or later.
  * https://github.com/confluentinc/librdkafka/pull/4353

### DIFF
--- a/fluent-package/yum/amazonlinux-2023/Dockerfile
+++ b/fluent-package/yum/amazonlinux-2023/Dockerfile
@@ -47,6 +47,8 @@ RUN \
     ruby3.2 \
     ruby3.2-rubygems \
     ruby3.2-rubygem-rake \
+    # install which for building librdkafka 1.9.0 or later
+    which \
     && \
   # raise IPv4 priority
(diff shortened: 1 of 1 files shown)


fluent-package/yum/amazonlinux-2023/Dockerfile
```diff
@@ -28,6 +28,7 @@ RUN \
   yum install -y ${quiet} yum-utils && \
   yum groupinstall -y ${quiet} "Development Tools" && \
   yum install -y ${quiet} \
+    which \
```
daipom, line 31:
```suggestion
    # install which for building librdkafka 1.9.0 or later 
    which \
```

How about adding a comment like the others?

https://github.com/fluent/fluent-package-builder/blob/dc81219b7174e0b8e1523643df383a1971fc286b/fluent-package/yum/amazonlinux-2/Dockerfile#L48-L49

kenhys (approved):
LGTM.

PR #786, opened 2025-02-03
Backport: amazonlinux-2023: ensure to install `which` command (#784)

kenhys:
Backport #784 

* It appears that the recent Amazon Linux 2023 image stops providing `which`.
* We need `which` to build `librdkafka 1.9.0` or later.
  * https://github.com/confluentinc/librdkafka/pull/4353


kenhys:
Test failure should be fixed in another PR.

PR #787, opened 2025-02-04
github: enable arm CI with ubuntu-24.04-arm host

kenhys:


kenhys:
Try whether building on arm host is reasonable or not on recent GitHub Actions.

kenhys:
It seems that surely amzn 2023 build takes around 20minues, but it has a merit that no need to build separately.

kenhys:
In the future, foo-arm.yml might be unified into one .yml, but it is out-of-scope in this PR.

kenhys:
TODO: backport to LTS.

fluent-package/apt/commonvar.sh
```diff
@@ -16,6 +16,14 @@ case ${code_name} in
...
     channel=universe
     mirror=http://archive.ubuntu.com/ubuntu/
+    case ${code_name} in
+        focal|jammy)
+            if [ "$architecture" = "arm64" ]; then
+                echo "For ${code_name} (arm64), use ubuntu-ports"
+                mirror=http://ports.ubuntu.com/ubuntu-ports
+            fi
+            ;;
+    esac
```
daipom, line 26: Could you tell me why we need this?

kenhys: For example, arm64 package is not available from archive.ubuntu.com.

https://github.com/fluent/fluent-package-builder/actions/runs/13127385504/job/36626186947
https://github.com/fluent/fluent-package-builder/actions/runs/13127385504/job/36626187253

kenhys: but it might be better not to switch code_name here.

kenhys: Fixed.
(temporary piuparts execution was disable for noble, but it is same for noble)

daipom: I see! Thanks!

daipom (approved):
LGTM. Thanks!

PR #789, opened 2025-02-05
Backport github: enable arm CI with ubuntu-24.04-arm host (#787)

kenhys:
Backport #787 for LTS branch.

In the previous versions, LTS package for arm was built on bare metal arm machine.
It might be better to build it with GitHub Actions to integrate them for LTS release process.


daipom (approved):
Thanks!

PR #794, opened 2025-02-10
rpm: keep compatibility with createrepo

kenhys:
Latest createrepo-c breaks compatibility with createrepo. For example, it does not create metadata (repodata/*) which can be handled yum in AmazonLinux 2.

To keep compatibility for such a distribution, need to specify --compatibility explicitly.

See https://github.com/rpm-software-management/createrepo_c/issues/383


Watson1978 (approved):
LGTM.
Thanks 👍🏻

daipom (approved):
Thanks!

PR #802, opened 2025-03-27
Revise index.html to browse them

kenhys:
* Use latest REMIX ICON (Apache License 2.0)
* Add navigate to folders
* Add size in bytes

### DIFF
--- a/fluent-package/make-index-html.erb
+++ b/fluent-package/make-index-html.erb
@@ -6,7 +6,7 @@
     <!-- https://github.com/franciscop/picnic MIT license -->
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/picnic">
     <!-- https://github.com/Remix-Design/RemixIcon - Apache 2.0 license -->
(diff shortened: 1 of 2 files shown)


kenhys:
For example, makes it easy to browse repodata/ for digging into the problematic repository metadata.

image

fluent-package/make-index-html.erb
```diff
@@ -26,15 +26,24 @@
        </a>
      </div>
      <div style="margin: 0em auto 0em 2em">
-       Index of <a href="index.html"><%= options[:site] %><%= relative_path %></a>
+       Index of <a href="index.html"><%= options[:site] %>/<%= relative_path %></a>
+     </div>
```
daipom, line 30:
```suggestion
```

Looks like this `</div>` is unintentional.

daipom (approved):
Thanks! LGTM!

fluent-package/make-index-html.erb
kenhys: Thanks, fixed it.

PR #803, opened 2025-03-31
ci: use Incus instead of LXD

kenhys:
It seems that AmazonLinux:2 container image is not available anymore from https://images.lxd.canonical.com/.
AmazonLinux:2 is still available from
https://images.linuxcontainers.org,
but it can't be added without interactive operation for LXD.

It seems that it is easy to use Incus as a workaround.

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -189,19 +189,19 @@ jobs:
         include:
           - label: Debian bullseye amd64
             rake-job: debian-bullseye
-            lxc-image: images:debian/11
+            container-image: images:debian/11
           - label: Debian bookworm amd64
             rake-job: debian-bookworm
-            lxc-image: images:debian/12
+            container-image: images:debian/12
           - label: Ubuntu Focal amd64
             rake-job: ubuntu-focal
-            lxc-image: ubuntu:20.04
+            container-image: images:ubuntu/20.04
           - label: Ubuntu Jammy amd64
             rake-job: ubuntu-jammy
-            lxc-image: ubuntu:22.04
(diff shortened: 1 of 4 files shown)


kenhys:
CI for windows failure is out of scope in this PR.

kenhys:
NOTE: CI will be broken on April 15. (CGroup v1)

daipom (approved):
Thanks!

Watson1978 (approved):
👍🏻

.github/workflows/apt.yml
```diff
@@ -235,24 +235,28 @@ jobs:
...
-          echo "::endgroup::"
-          echo "::group::lxc remote list"
-          lxc remote list
-          echo "::endgroup::"
-          echo "::group::lxc list images:"
-          lxc image list images:
-          echo "::endgroup::"
-      - name: Run Test  ${{ matrix.test }} on ${{ matrix.lxc-image }}
-        run: fluent-package/apt/systemd-test/test.sh ${{ matrix.lxc-image }} ${{ matrix.test }}
+          sudo curl -fsSL https://pkgs.zabbly.com/key.asc -o /etc/apt/keyrings/zabbly.asc
```
komainu8, line 240: If we use ubuntu 24.04, incus include Ubuntu official repository.
So, we don't use pkgs.zabbly.com.

See:
https://github.com/mroonga/mroonga/blob/main/.github/workflows/linux.yml#L275-L278
https://packages.ubuntu.com/noble/incus

kenhys: Oh, thanks!

PR #804, opened 2025-04-03
ci: fix build error on Windows

Watson1978:
To build the cmetrics gem with CMake 4.0, `CMAKE_POLICY_VERSION_MINIMUM` should be set appropriately, like:

```
      env:
        # Required to build cmetrics gem with CMake 4.0+
        CMAKE_POLICY_VERSION_MINIMUM: 3.5
```

Ref. https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_VERSION_MINIMUM.html

We are using docker to create Windows packages, but it seems that the environment variables are not propagating properly there.

So, I did the downgrade the CMake version.


kenhys (approved):
LGTM.

It is reasonable.

kenhys:
Thanks!

PR #806, opened 2025-04-04
ci: bump hostedtoolcache Ruby to 3.2

kenhys:
Upcoming Fluentd requires Ruby 3.2 or later.
It fixes the following error:

```
  stderr: fluentd-1.18.0 requires ruby version >= 3.2, which is
  incompatible with the current version, 3.1.7
```


kenhys:
Extracted commit from WIP #805

Watson1978 (approved):
👍🏻

PR #807, opened 2025-04-07
ci: switch CGroup v1 test with VirtualBox

kenhys:
ci: switch CGroup v1 test on vagrant (VirtualBox)

In the previous versions, AmazonLinux:2 image
was lost from LXD, so switched from LXD to Incus.

In this time, Ubuntu 20.04 runner will be removed from
GitHub Actions, so need to give up using 20.04 runner
for upcoming EOL (2025/04/15).
Thus, there is no CGroup V1 host runner available on GitHub Actions.
As a result, it means that there is no straightforward way to test package
on CGroup v1 host. (such as AmazonLinux:2, AmazonLinux:2023)

Instead, use vagrant (VirtualBox) to test fluent-package.
It enables to run VM with CGroup v1 to keep CI alive.

### DIFF
--- a/.github/workflows/yum.yml
+++ b/.github/workflows/yum.yml
@@ -130,7 +130,10 @@ jobs:
   v1test:
     name: Test ${{ matrix.label }} ${{ matrix.test }} (CGroup V1)
     needs: build
-    runs-on: ubuntu-20.04
+    # Ubuntu 20.04 is not available anymore, so can't use container based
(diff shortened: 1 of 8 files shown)


daipom (approved):
Thanks so much!!

daipom:
Note: We should unify `apt-get` / `apt`. Maybe in another PR.

PR #808, opened 2025-04-07
ci: skip test when cache is enabled

kenhys:
It omits redundant tests.

* Usually source code was changed, and always ci tests them. It might be redundant when it will be tested with subsequent commits during development.
* When they were cached, ci skip tests. It assumes that CI pass test case in the previous commit
   * If you want to ensure CI results in that commit again, invoke workflow dispatch.

### DIFF
--- a/.github/workflows/apt-arm.yml
+++ b/.github/workflows/apt-arm.yml
@@ -5,6 +5,7 @@ on:
       - master
       - fluent-package-v5
   pull_request:
+  workflow_dispatch:
 concurrency:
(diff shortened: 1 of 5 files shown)


Watson1978 (approved):
👍🏻

daipom:
I understand that we want to skip the build if possible, since it takes a lot of time.
However, I think it’s fine to run the tests every time, even if it feels redundant.

kenhys:
It might be good for the development phase, but not for production, so I'll close it.

PR #809, opened 2025-04-08
ci: split some binary package tests

kenhys:
In the previous versions, some test cases are contained in Build job.
It should be executed in a separated job by nature.


Watson1978 (approved):
👍🏻

PR #810, opened 2025-04-10
deb: refresh target platform

kenhys:
* Add debian:trixie
* Drop debian:bullseye
* Drop ubuntu:focal

Make forward for upcoming v6 series.


daipom (approved):
Thanks! LGTM!

kenhys:
thanks.

PR #811, opened 2025-04-14
ci: make cache key unique to reuse cache correctly

kenhys:
Unexpectedly, cache key was same with amd64/x86_64 workflows. It should be fixed with adding -arm64/-aarch64 identifier.

Before:
* amd64/x86_64 and arm64/aarch64 packages were cached in same key
After: 
* amd64/x86_64 and arm64/aarch64 packages were cached separately 
* It reduces each cache size

### DIFF
--- a/.github/workflows/apt-arm.yml
+++ b/.github/workflows/apt-arm.yml
@@ -49,7 +49,7 @@ jobs:
         with:
           path: |
             fluent-package/apt/repositories
-          key: ${{ runner.os }}-cache-${{ matrix.rake-job }}-${{ hashFiles('**/config.rb', '**/Rakefile', '**/Gemfile*', 'fluent-package/templates/**', 'fluent-package/debian/**', 'fluent-package/apt/**/Dockerfile') }}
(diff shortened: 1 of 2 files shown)


kenhys:
TODO: rebase with #810

kenhys:
rebased with master

daipom:
Thanks! 
Looks like it was wrong since

* #787

I have noticed there is no `if: ${{ ! steps.cache-deb.outputs.cache-hit }} ` condition to build.
Do we need it as well?

Ref:

https://github.com/fluent/fluent-package-builder/blob/cff71d1c0d75a4bd97eee91564eb3a60c3f46726/.github/workflows/apt.yml#L48-L63

daipom:
> I have noticed there is no if: ${{ ! steps.cache-deb.outputs.cache-hit }} condition to build.

Oh, sorry, #812 fixes it.

daipom (approved):
LGTM. Thanks!

Watson1978 (approved):
👍🏻

PR #814, opened 2025-04-15
Add support for Linux capability

kenhys:
Linux capability was documented in https://docs.fluentd.org/deployment/linux-capability, but capng_c was not bundled yet.


daipom (approved):
Thanks! Now we can use `fluent-cap-ctl` by default!

Watson1978 (approved):
👍🏻

PR #818, opened 2025-05-07
Backport: ci: use Incus instead of LXD (#803)

kenhys:
Backport #803 

It seems that AmazonLinux:2 container image is not available anymore from https://images.lxd.canonical.com/.
AmazonLinux:2 is still available from
https://images.linuxcontainers.org,
but it can't be added without interactive operation for LXD.

It seems that it is easy to use Incus as a workaround.


Watson1978 (approved):
👍🏻

PR #819, opened 2025-05-07
Backport ci: bump hostedtoolcache Ruby to 3.2 (#806)

kenhys:
Backport #806 

Upcoming Fluentd requires Ruby 3.2 or later.
It fixes the following error:

```
  stderr: fluentd-1.18.0 requires ruby version >= 3.2, which is
  incompatible with the current version, 3.1.7
```

NOTE: Though v1.18 will not be shipped for v5 branch, it is better not to use EOL toolchain.


Watson1978 (approved):
👍🏻

PR #812, opened 2025-04-14
ci: skip "Install dependencies" conditionally

kenhys:
Before:
 Always execute "Install dependencies" even though build step
 could be skipped.

After:
 Skip "Install dependencies" step if build step is omitted.

It will reduces workflow execution times.

PR #825, opened 2025-05-08
Backport ci: skip "Install dependencies" conditionally (#812)

kenhys:
Before:
 Always execute "Install dependencies" even though build step
 could be skipped.

After:
 Skip "Install dependencies" step if build step is omitted.

It will reduces workflow execution times.


kenhys:
It assumes backport #824

kenhys:
rebased with fluent-package-v5 branch

Watson1978 (approved):

PR #827, opened 2025-05-09
ci: extract setup rpmrebuild and fix rebuild

kenhys:
RpmUnpack implementation was changed from rpm2cpio to rpm2archive, but it can't rebuild RPM because it handles outcome of rpm2archive wrongly. It will be fixed in rpmrebuild 2.21.

See
  rpmrebuild 2.20: rpm2archive called with incorrect arguments
  https://sourceforge.net/p/rpmrebuild/bugs/18/

### DIFF
--- a/fluent-package/yum/systemd-test/setup-rpmrebuild.sh
+++ b/fluent-package/yum/systemd-test/setup-rpmrebuild.sh
@@ -1,0 +1,28 @@
+case $distribution in
+    amazon)
+        case $version in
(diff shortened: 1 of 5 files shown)


kenhys:
checked with https://github.com/kenhys/fluent-package-builder/actions/runs/14919976707
(reduced target AlmaLinux9 only)

daipom (approved):
LGTM! Thanks!

PR #838, opened 2025-05-19
ci: apply bundle-audit

kenhys:
Not to let vulnerability alone, use bundle-audit


kenhys:
If need to ignore specific warning, put .bundler-audit.yml in each branch.

Watson1978 (approved):

PR #840, opened 2025-05-20
package: navigate with index.html explicitly

kenhys:
On object storage, directory index is not available, so use generated DIRECTORY/index.html explicitly.


kenhys:
Reproduced:

https://fluentd.cdn.cncf.io/lts/5/index.html

daipom (approved):
Thanks! LGTM.

PR #841, opened 2025-05-22
ci: fix wrong grouping about CGroup v1

kenhys:
AmazonLinux 2023 contains code for CGroup v1, but it is not officially supported.

See https://docs.aws.amazon.com/eks/latest/userguide/al2023.html

    While AL2023 still includes code that can make the system run using
    cgroupv1, this isn’t a recommended or supported configuration. This
    configuration will be completely removed in a future major release of
    Amazon Linux.

### DIFF
--- a/.github/workflows/yum.yml
+++ b/.github/workflows/yum.yml
@@ -195,7 +195,6 @@ jobs:
       matrix:
         label:
           - AmazonLinux 2 x86_64
-          - AmazonLinux 2023 x86_64
         test:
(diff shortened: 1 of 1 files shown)


kenhys:
Assume that AL2023 test case will pass with this PR.

kenhys:
As #842 was merged into master, so rebase it.

kenhys:
CGroup v1 target is now reduced to AL2.

daipom (approved):
Thanks!

PR #847, opened 2025-05-28
msi: recommend Fluentd enterprise service

kenhys:
In the previous versions, there were no navigation to access Fluentd enterprise services.

It might be useful to provide such navigation link for Fluentd users.


kenhys:
The following toast message will be shown.

fluent-package-toast-screenshot

kenhys:
If "access website" checkbox is implemented in custom page, this feature (toast) might be redundant.

Watson1978 (approved):
👍🏻

kenhys:
Let's merge when ack from @daipom .

PR #854, opened 2025-06-19
deb: use appropriate separator for lintian

kenhys:
It fixes the following error:

```
  Please use spaces as separators in field Disable-Tags instead of commas in profile /root/.lintian/profiles/fluent-package/ubuntu.profile
  Error: The operation was canceled.
```


kenhys:
I guess it was caused newer version of  lintian on jammy and noble, but not confirmed

kenhys:
> Do we need to backport this?

Though I haven't checked with LTS branch yet, I think so.

PR #861, opened 2025-07-23
config: update fluentd revision

Watson1978:


kenhys:
Let's merge when the rest of workflow succeeds.

kenhys:
workflow seems stalled.

PR #870, opened 2025-07-31
follow to verify with v6 experimental package

kenhys:
Now packages for verification is in-progress,
it should follow test/experimental/6.


Watson1978 (approved):
👍🏻

PR #873, opened 2025-08-05
Backport: switch from treasuredata.com to fluentd.cdn.cncf.io

kenhys:


kenhys:
It seems that Focal image is not available...

daipom (approved):
Thanks!

PR #875, opened 2025-08-12
Add builtin config files

Watson1978:
* Add config file for fluent-plugin-obsolete-plugins
* Add config file for fluent-plugin-fluent-package-update-notifier


.github/workflows/yum.yml
```diff
@@ -301,6 +301,7 @@ jobs:
           - "update-to-next-major-version.sh etc inactive"
           - "update-without-data-lost.sh v6 v7"
           - "update-without-data-lost.sh v7 v6"
+          - "builtin-plugins.sh"
```
kenhys, line 304: As take it into account for adding update-notifier test case in the future, 
obsolete-plugins.sh might be better to clarify.

kenhys (changes requested):

kenhys (approved):
LGTM.

kenhys:
Already similar check has passed.

Watson1978:
Thanks

PR #879, opened 2025-08-22
ci: update Ruby version in maintenance workflow

Watson1978:
I forgot to update Ruby version in `maintenance-bundle-update.yml`

It should have same Ruby version in 
https://github.com/fluent/fluent-package-builder/blob/401960014daeb15ede44fcdc76ee373d65ffce34/fluent-package/config.rb#L26


kenhys (approved):
Good catch.

PR #885, opened 2025-08-25
ci: code refactoring for reuse for RPM

daipom:
Code for installing RPM is complex, so we should reuse it for managing the test code.


daipom:
All related tests passed

kenhys (changes requested):
Good catch!

I'll agree with these code refactoring, but

a) rename file name for suitable one 
  or
b) putting these snippet into isolated file might be better. (not common variables)

daipom:
Thanks!

PR #888, opened 2025-08-26
Improve utility script not to fail with empty response

kenhys:
There is a case that the response of API doesn't match expected pull request.
As a result, it must follow the next one or more to search.

### DIFF
--- a/fluent-package/manage-fluent-repositories.sh
+++ b/fluent-package/manage-fluent-repositories.sh
@@ -199,12 +199,22 @@ EOF
 	head_branch=$(echo $response | cut -d' ' -f1)
 	head_sha=$(echo $response | cut -d' ' -f2)
 	rm -f dl.list
-	curl --silent --location \
-	     -H "Accept: application/vnd.github+json" \
-	     -H "Authorization: Bearer $GITHUB_ACCESS_TOKEN" \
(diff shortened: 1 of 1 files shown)


Watson1978 (approved):

fluent-package/manage-fluent-repositories.sh
```diff
@@ -197,12 +199,22 @@ EOF
 	head_branch=$(echo $response | cut -d' ' -f1)
 	head_sha=$(echo $response | cut -d' ' -f2)
 	rm -f dl.list
-	curl --silent --location \
-	     -H "Accept: application/vnd.github+json" \
-	     -H "Authorization: Bearer $GITHUB_ACCESS_TOKEN" \
-	     -H "X-GitHub-Api-Version: 2022-11-28" \
-	     "https://api.github.com/repos/fluent/fluent-package-builder/actions/artifacts?per_page=100&page=$d" | \
```
daipom, line 204: Hmm, what was this original `$d`? Just a typo or something?

kenhys: I guess it was typo, but harmless because it was not actually used.

daipom: I see!

PR #889, opened 2025-08-26
rpm: pack fluent-lts-release as artifact

kenhys:
NOTE: when beginning LTS branch fluent-package-v6, stop to ship fluent-lts-release from master branch.


daipom (approved):
Thanks!

PR #890, opened 2025-08-27
rpm: support to convert fluent-release package too

kenhys:
Recently fluent-release package was introduced, to convey them correctly from artifacts to our 
repository layout, need to support it explicitly.


daipom (approved):
Thanks!

PR #892, opened 2025-08-27
ci: add AlmaLinux 10 for RPM tests

daipom:


daipom:
> It appears that we need to resolve rpmrebuild installation issue on AlmaLinux 10.

Rebasing on the following fix and excluding the failing tests would be one solution.

* #897

PR #896, opened 2025-08-27
yum: fix target artifact name

kenhys:
It should be without -lts- for fluent-release package


kenhys:
TODO: need to tweak

kenhys:
It might be better to use unique artifact name between master (v6) and v5.

daipom (approved):
Thanks!

Note: We will manage `fluent-release` for v5 normal channel on this branch exceptionally (although it is not related to v5 LTS)

PR #897, opened 2025-08-27
ci: reinforce update-to-next-major-version test

daipom:
The new test code are from other update-to-next-version tests.
Those tests need rpmrebuild, but looks like it is not supported on EPEL for AlmaLinux 10 yet.
So, those tests can not be done for AlmaLinux 10.
It would be better to cover those tests on this test as well.


daipom:
This would be one solution for #892.

PR #907, opened 2025-09-14
deb: add missing Architectures: field

kenhys:
It is better to specify supported architectures in Architectures: field explicitly.

Without it, if other architecture was enabled (e.g. i386 was enabled additionally with multiarch), apt emits the following message:

```
  Notice: Skipping acquire of configured file
  'contrib/binary-i386/Packages' as repository
  'https://fluentd.cdn.cncf.io/lts/6/debian/trixie trixie InRelease'
  doesn't support architecture 'i386'
```


daipom (approved):
Thanks!

PR #908, opened 2025-09-16
add security policy

daipom:
Intended to be used in the following fix:

* https://github.com/fluent/fluentd/pull/5098


kenhys (approved):
LGTM.

PR #910, opened 2025-09-19
packages: add helper script to update gemfile

kenhys:
For LTS channel:

$ ruby update-gemfile.rb --lts

For standard channel:

$ ruby update-gemfile.rb


kenhys:
e.g. for LTS

image

PR #915, opened 2025-09-22
deb: try to use custom fluent image

kenhys:
images:ubuntu/20.04 is not available anymore, so
trying to use custom Incus image.


Watson1978 (approved):
👍🏻

Issue #590, opened 2023-09-22
.package_note-fluent-package-5.0.1-1.amzn2023.x86_64.ld is no exist


wdxal:
OS:AmaonLinux2023
Version:5.0.1-1(LTS)
Install:
Step 1:
`curl -fsSL https://toolbelt.treasuredata.com/sh/install-amazon2023-fluent-package5-lts.sh | sh`
Step 2:
`fluent-gem install mysql2`

Error:
`$ cat /opt/fluent/lib/ruby/gems/3.2.0/extensions/x86_64-linux/3.2.0/mysql2-0.5.5/mkmf.log`
LD_LIBRARY_PATH=.:/opt/fluent/lib "gcc -o conftest -I/opt/fluent/include/ruby-3.2.0/x86_64-linux -I/opt/fluent/include/ruby-3.2.0/ruby/backward -I/opt/fluent/include/ruby-3.2.0 -I.    -O2 -ftree-vectorize -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -march=x86-64-v2 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -fPIC conftest.c  -L. -L/opt/fluent/lib -Wl,-rpath,/opt/fluent/lib -L. -Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -Wl,--build-id=sha1 -Wl,-dT,/root/rpmbuild/BUILD/fluent-package-5.0.1/.package_note-fluent-package-5.0.1-1.amzn2023.x86_64.ld -fstack-protector-strong -rdynamic -Wl,-export-dynamic -Wl,--no-as-needed  -m64   -Wl,-rpath,/opt/fluent/lib -L/opt/fluent/lib -lruby  -lm -lpthread  -lc"
[...]

PR #877, opened 2025-08-13
yum: Disable package-note-file to remove unnecessary linker flag

Watson1978:
The user can refer the Ruby configure results through `RbConfig::CONFIG`.
`RbConfig::CONFIG` is mainly used when installing gems which has C extension.

If flags remain that use linker scripts that do not exist in the user's environment,
it causes errors when user install the gem.

When creating a package using `fluent-package.spec.in`, the environment variable `LDFLAGS` was automatically set as follows.

```
LDFLAGS='-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1    -Wl,--build-id=sha1 -Wl,-dT,/root/rpmbuild/BUILD/fluent-package-6.0.0/.package_note-fluent-package-6.0.0-1.amzn2023.x86_64.ld'
```

Ruby's `configure` refers to the environment variable `LDFLAGS`,
and it expands into `RbConfig::CONFIG`.

Fix https://github.com/fluent-plugins-nursery/fluent-plugin-node-exporter-metrics/issues/21
Fix https://github.com/fluent/fluent-package-builder/issues/590

PR #916, opened 2025-09-24
Backport(v5): yum: Disable package-note-file to remove unnecessary linker flag (#877)

Watson1978:
Backport #877

The user can refer the Ruby configure results through `RbConfig::CONFIG`.
`RbConfig::CONFIG` is mainly used when installing gems which has C extension.

If flags remain that use linker scripts that do not exist in the user's environment,
it causes errors when user install the gem.

When creating a package using `fluent-package.spec.in`, the environment variable `LDFLAGS` was automatically set as follows.

```
LDFLAGS='-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1    -Wl,--build-id=sha1 -Wl,-dT,/root/rpmbuild/BUILD/fluent-package-6.0.0/.package_note-fluent-package-6.0.0-1.amzn2023.x86_64.ld'
```

Ruby's `configure` refers to the environment variable `LDFLAGS`, and it expands into `RbConfig::CONFIG`.

Fix
https://github.com/fluent-plugins-nursery/fluent-plugin-node-exporter-metrics/issues/21 Fix https://github.com/fluent/fluent-package-builder/issues/590

---------

### DIFF
--- a/fluent-package/yum/fluent-package.spec.in
+++ b/fluent-package/yum/fluent-package.spec.in
@@ -138,6 +138,13 @@ The stable distribution of Fluentd, formerly known as td-agent.
 %if %{use_scl_ruby}
 . /opt/rh/rh-ruby%{scl_ruby_ver}/enable
 %endif
+
+# Disable package-note-file which embeds build environment path
+# in LDFLAGS. It blocks build process with third-party gems which
+# requires compile procedure.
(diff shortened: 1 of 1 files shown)


kenhys (approved):
LGTM.

PR #921, opened 2025-10-02
tmpfiles.d: ensure tmp dir for lock files not be cleaned up

daipom:
We should exclude `/tmp/fluentd-lock-*` directory.
If `/tmp/fluentd-lock-*` is cleaned up while Fluentd is running:

* out_file can fail flushing.
* out_secondary can fail flushing.
* The supervisor process exit abnormally when Fluentd stops.

At least, RHEL 8 has the following default conf for tmpfiles.d:

/usr/lib/tmpfiles.d/tmp.conf

    q /tmp 1777 root root 10d

This can delete directories under `/tmp` if they are not used for more than 10 days.

We don't need to protect the old lock files under
`/tmp/fluentd-lock-*/`, so use `X` instead of `x`.

In configs where lock files are not used, there is no problem even if this directory is deleted.
In configurations that do use lock files, it’s unlikely that this directory would not be updated for more than 10 days. Therefore, the cases where this actually becomes an issue should be quite limited.
If only out_secondary_file uses lock files, this can become an issue.

### DIFF
--- a/.github/workflows/apt.yml
+++ b/.github/workflows/apt.yml
@@ -253,6 +253,7 @@ jobs:
           - "update-without-data-lost.sh v6 v7"
           - "update-without-data-lost.sh v7 v6"
           - "obsolete-plugins.sh"
+          - "tmpfiles.sh"
         include:
           - label: Debian bookworm amd64
             rake-job: debian-bookworm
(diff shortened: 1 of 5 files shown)


Watson1978 (approved):
👍🏻

PR #855, opened 2025-07-09
Backport: Add fluent-release package for yum/dnf update (#845)

kenhys:
* Backport #845

rpm: add fluent-release package for yum/dnf update

In the previous versions, .repo file was assumed to be created via fluent-package installation script.

For migration from packages.treasuredata.com to fluentd.cdn.cncf.io in the future, it might be better to setup proper fluent-release package for it.

Changed in backport

* baseurl was changed to LTS
* fluent-package.repo => fluent-package-lts.repo
* changed section name (fluent-package-lts => fluent-package-lts-v5) in fluent-package-lts.repo
* fluent-package-lts-v5 section in fluent-package-lts.repo was enabled by default

PR #923, opened 2025-10-02
v5: ci: fix backport error

Watson1978:
This PR will fix backport error in #917

When I solve the conflict, I introduced the error.
Therefore, the `yum` workflow does not works properly.


Watson1978:
I think this will fix error in https://github.com/fluent/fluent-package-builder/actions/runs/18188027315

image

daipom:
We also should address the following.

https://github.com/fluent/fluent-package-builder/blob/0f6e3e05cb05c0be1b571e8bc23826c7f061bbca/.github/workflows/yum.yml#L266-L269

The backport #855 added a broken duplicated step.
I have added the commit.

Watson1978:
> The backport #855 added a broken duplicated step.
> I have added the commit.

Thanks.

daipom (approved):
Thanks!

PR #924, opened 2025-10-03
Backport(v5): tmpfiles.d: ensure tmp dir for lock files not be cleaned up (#921)

Watson1978:
Backport #921

We should exclude `/tmp/fluentd-lock-*` directory. If `/tmp/fluentd-lock-*` is cleaned up while Fluentd is running:

* out_file can fail flushing.
* out_secondary can fail flushing.
* The supervisor process exit abnormally when Fluentd stops.

At least, RHEL 8 has the following default conf for tmpfiles.d:

/usr/lib/tmpfiles.d/tmp.conf

    q /tmp 1777 root root 10d

This can delete directories under `/tmp` if they are not used for more than 10 days.

We don't need to protect the old lock files under
`/tmp/fluentd-lock-*/`, so use `X` instead of `x`.

In configs where lock files are not used, there is no problem even if this directory is deleted.
In configurations that do use lock files, it’s unlikely that this directory would not be updated for more than 10 days. Therefore, the cases where this actually becomes an issue should be quite limited. If only out_secondary_file uses lock files, this can become an issue.

### DIFF
--- a/fluent-package/apt/systemd-test/tmpfiles.sh
+++ b/fluent-package/apt/systemd-test/tmpfiles.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+set -exu
+
+. $(dirname $0)/../commonvar.sh
+
+# Display unit info for debug
+sudo systemctl cat systemd-tmpfiles-clean.service
+sudo systemctl cat systemd-tmpfiles-clean.timer
+
+# Install the built package
+sudo apt install -V -y \
+    /host/${distribution}/pool/${code_name}/${channel}/*/*/fluent-package_*_${architecture}.deb
+
+# Wait all processes to start
+systemctl status --no-pager fluentd
+sleep 3
+
+# Test: the files under /tmp/ exist and not be cleaned up by default
+ls -d /tmp/fluent
+ls -d /tmp/fluentd-lock-*
+sudo systemd-tmpfiles --clean
+ls -d /tmp/fluent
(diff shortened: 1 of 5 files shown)


fluent-package/apt/systemd-test/tmpfiles.sh
daipom, line 1: We can't use the functions in `commonvar.sh` since they are not backported. :cry: 

https://github.com/fluent/fluent-package-builder/blob/dd3921119bfad19daf73a13a33d8c4b57361ce51/fluent-package/yum/systemd-test/common.sh#L38-L41

fluent-package/yum/systemd-test/tmpfiles.sh
daipom, line 1: We can't use the functions in `common.sh` since they are not backported. :cry: 

https://github.com/fluent/fluent-package-builder/blob/dd3921119bfad19daf73a13a33d8c4b57361ce51/fluent-package/yum/systemd-test/common.sh#L38-L41

Watson1978: Oops

Watson1978: Fixed with https://github.com/fluent/fluent-package-builder/pull/924/commits/35746b2f7830518568b90caa8330007a58475c54

daipom (approved):
Thanks!

PR #927, opened 2025-10-08
msi: fix bug where install fails for path with spaces

### LINKED ISSUE #926: fluent-package-6.0.0-x64.msi cannot be installed to folder with spaces on Windows
Problem
fluent-package-6.0.0-x64.msi cannot be installed to a folder with spaces on windows
The application is being isntalled via msi wizard. It is allowed to choose any directory (for example c:\Program Files). While installation the installer rollbacks back leaving corruputed system: some registry keys remain and blocking other tries to install (to default folder or any other without spaces).
Machine reboot doesn't help until registry manual cleaning up is performed
No official documentation is found regarding any limitations for installation folder or possible work arounds

OS used for installation: Windows 10 Pro, 64 bit

Expected
User can select any folder to install fluent-package.

daipom:
* Fixes #926

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -26,7 +26,7 @@ jobs:
         id: cache-msi
         with:
           path: fluent-package/msi/repositories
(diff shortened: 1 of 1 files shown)


daipom:
Checking CI fails without the fix.
I will add the fix later.

PR #982, opened 2026-02-26
msi: change ExecutionPolicy to enable it

kenhys:
In the previous versions, when RemoteSigned was Restricted, it fails
    to execute toast.ps1, thus it causes installation rollback unexpectedly.
    So, it should specify policy of powershell explicitly

Even though execution was disabled via GPO, it just fails but
    always fallback to exit /b 0.


Watson1978 (approved):
Thanks 👍🏻

PR #1016, opened 2026-03-28
ci: enable ubuntu 26.04 snapshot for testing

kenhys:
use fluent:ubuntu/26.04 snapshot in advance for a while.
That will be replaced with linuxcontainer's image later.


Watson1978 (approved):
👍🏻

PR #1033, opened 2026-04-30
Backport(v6) ci deb: use reusable workflow (#1032)

kenhys:
In the previous versions, current (v6) and next major (v7) version were built in same workflow job.
It tend to reach maximum timeout (6h) easily in recent actions. So let's split into two jobs and use reusable workflow for them.

* previous build job was split into build, build-next
* to reuse workflow for amd64 and arm64, runner.arch to distinct them
* v7 could be buildable for arm64 now (disabled)


Watson1978 (approved):
👍🏻

PR #1042, opened 2026-05-07
ci: use linuxcontainers image

kenhys:
fluent:ubuntu/26.04 is snapshot image because 26.04 is not ready at that time.

Now resolute image was officially supported, switch to it.


Watson1978 (approved):
👍🏻

PR #1056, opened 2026-05-29
github: fix broken minutes format

kenhys:
Fix broken format which was introduced by #964.

Before: -HHMM means hour-month
After: -HHmmss means hour-minutes-seconds

By adding "seconds" in format, avoid conflict by
sequentially fired actions.


Watson1978 (approved):
👍🏻

PR #1057, opened 2026-06-02
ci: use confluent platform 7.9

kenhys:
EOL: Feb 19, 2027

https://docs.confluent.io/platform/current/installation/versions-interoperability.html#cp-and-apache-ak-compatibility


Watson1978 (approved):
👍🏻

PR #1062, opened 2026-06-04
ci: fixup broken mirrors

kenhys:
There is a case that mirrorlist is not available
form master mirror site.
This is rare case, but it happens.

https://github.com/fluent/fluent-package-builder/actions/runs/26812024181/job/79058336020?pr=1028 https://github.com/fluent/fluent-package-builder/actions/runs/26812024181/job/79058337196?pr=1028

PR #1063, opened 2026-06-05
Backport(v6) ci: fixup broken mirrors (#1062,#1064)

kenhys:
There is a case that mirrorlist is not available
form master mirror site.
This is rare case, but it happens.

https://github.com/fluent/fluent-package-builder/actions/runs/26812024181/job/79058336020?pr=1028 
https://github.com/fluent/fluent-package-builder/actions/runs/26812024181/job/79058337196?pr=1028

PR #1064, opened 2026-06-05
ci: ensure ready to use config-manager

kenhys:
Follow-up #1062

There was a case that config-manager is not instaled.

https://github.com/fluent/fluent-package-builder/actions/runs/26989986882/job/79648097423


kenhys:
changed to install dnf-plugins-core in advance.

https://github.com/fluent/fluent-package-builder/actions/runs/26990932821/job/79650861124?pr=1064

kenhys:
merge into #1063 later.

Watson1978 (approved):
👍🏻

PR #1068, opened 2026-06-18
ci: force newer CMake policy

kenhys:
newer CMake drops policy version < 3.5, so keeping build with older policy (e.g. cmetrics), raise border to 3.5 on windows-latest

```
  -----
  CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
    Compatibility with CMake < 3.5 has been removed from CMake.
    Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
    to tell CMake that the project requires at least <min> but has been updated
    to work with policies introduced by <max> or earlier.
    Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
  -- Configuring incomplete, errors occurred!
  ----- end of file -----
  *** extconf.rb failed ***
```

### DIFF
--- a/.github/workflows/maintenance-bundle-update.yml
+++ b/.github/workflows/maintenance-bundle-update.yml
@@ -19,6 +19,8 @@ jobs:
           ruby-version: 3.4 # Use same Ruby version with fluent-package bundled
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Bundle Update
+        env:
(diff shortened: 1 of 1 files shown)


kenhys:
#1069 was landed, so it is not requried for now.
But in the future, we need to follow newer CMake or force policy version.

kenhys:
Keep this PR open.

kenhys:
superseded by #1110

PR #1069, opened 2026-06-18
ci: switch back windows 2022

kenhys:
windows-latest was changed to windows-2025, in contrast to that, we use windows-2022 explicitly for windows CI.

Keep it windows-2022 for a while to update Gemfile.lock

(toolchan on windows-2025 was upgraded, it is not confirmed yet.)


Watson1978 (approved):
👍🏻

PR #1072, opened 2026-06-18
ci: introduce retry job to make it stable

kenhys:
There is a case that workflow jobs are failed time to time.
Reduce manual retry operation, add workflow for it.

NOTE: only execute retry twice.

https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_run

### DIFF
--- a/.github/workflows/retry-workflows.yml
+++ b/.github/workflows/retry-workflows.yml
@@ -1,0 +1,35 @@
+name: Retry Workflows
+on:
+  workflow_run:
+    workflows:
+      - Apt based Linux
+      - Apt based Linux (AArch64)
+      - Yum based Linux
+      - Yum based Linux (AArch64)
+      - Windows
+    types:
+      - completed
+
+permissions:
+  actions: write
+  contents: read
+
+jobs:
+  retry:
+    # Retry only twice
+    if: >-
+      contains(
+        fromJSON('["failure", "timed_out"]'),
+        github.event.workflow_run.conclusion
+      ) && github.event.workflow_run.run_attempt < 3
+    runs-on: ubuntu-latest
+    steps:
+      - name: Re-run failed jobs
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+          RUN_ATTEMPT: ${{ github.event.workflow_run.run_attempt }}
+        run: |
+          echo "(${RUN_ATTEMPT}/2) Re-running failed jobs for run ${RUN_ID}"
+          gh run rerun "${RUN_ID}" --failed


kenhys:
sometimes test job fails, try to re run automatically.

kenhys:
Avoid infinite loop, limit with run_attempt `if: failure() && github.run_attempt == 1`

kenhys:
changed to kick from another workflow.

.github/workflows/retry-workflows.yml
```diff
@@ -0,0 +1,39 @@
...
+  contents: read
+
+jobs:
+  retry:
+    # Retry only twice: first run + 2 retry attempt
+    if: >-
+      contains(
+        fromJSON('["failure", "timed_out"]'),
+        github.event.workflow_run.conclusion
+      ) && github.event.workflow_run.run_attempt < 4
```
Watson1978: The `run_attempt < 4` guard doesn't match the comment (`Retry only twice`) or the echo `(/3)`.
For 2 retries as intended, use `< 3`:

.github/workflows/retry-workflows.yml
```diff
@@ -0,0 +1,39 @@
...
+      - Apt based Linux (AArch64)
+      - Yum based Linux
+      - Yum based Linux (AArch64)
+      - Windows
+    types:
+      - completed
+
+concurrency:
+  group: ${{ github.head_ref || github.sha }}-${{ github.workflow }}
+  cancel-in-progress: true
```
Watson1978: The `concurrency.group` doesn't work as intended under `workflow_run`.

`github.head_ref` is only set for `pull_request` / `pull_request_target` events, so under `workflow_run` it's empty and the group always falls back to `github.sha`. And for `workflow_run`, `github.sha` is the last commit on the default branch, not the triggering run's commit.
Combined with the constant `github.workflow` (`Retry Workflows`), the group collapses to essentially one shared key across almost all retry runs.

Refs:
- `github.head_ref` is only set for `pull_request`/`pull_request_target`: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context
- `workflow_run` → `GITHUB_SHA` = "Last commit on default branch", `GITHUB_REF` = "Default branch": https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run

With `cancel-in-progress: true`, if two target workflows fail on the same commit around the same time (e.g. Yum and Windows), the later retry run cancels the earlier one. If that cancellation lands before `gh run rerun` fires, the failed workflow never gets retried — the opposite of what this PR is trying to achieve.

Keying the group on the triggering run makes it unique:

```yaml
concurrency:
  group: retry-${{ github.event.workflow_run.id }}-${{ github.event.workflow_run.run_attempt }}
  cancel-in-progress: false
```

Alternatively, since each `completed` event is already unique, the `concurrency` block could just be dropped.

.github/workflows/retry-workflows.yml
kenhys: Thanks, fixed.

.github/workflows/retry-workflows.yml
kenhys: Thanks, dropped it.

Watson1978 (approved):
👍🏻

PR #1078, opened 2026-06-26
msi: fix broken link to enterprise services

kenhys:


kenhys:
This is not applied to v6.0.4. Let's ship for v6.0.5.

Watson1978 (approved):

PR #1087, opened 2026-07-09
yum: keep compatibility with older RHEL 9.x

kenhys:
When building with latest RHEL 9.x, it breaks ABI compatibility with older RHEL 9.x because it requires GLIBC_2.35 and OPENSSL_3.4.0.

This is because glibc 2.35 symbols were backported into its glibc 2.34. It doesn't exist on RHEL 9.6 and earlier version.

Thus build environment should be pinned to 9.2 for keeping compatibility
with all range of RHEL 9.x.

See "Werk #18941: Fix installation failure on AlmaLinux 9 due to GLIBC_2.35 symbol mismatch"
https://checkmk.com/werk/18941

### DIFF
--- a/.github/workflows/_build_rpm_reusable.yml
+++ b/.github/workflows/_build_rpm_reusable.yml
@@ -42,6 +42,16 @@ jobs:
         if: ${{ ! steps.cache-rpm.outputs.cache-hit }}
         run: |
           rake yum:build YUM_TARGETS=${{ inputs.rake-job }}
+      - name: Ensure RHEL9 compatibility
+        if: ${{ ! steps.cache-rpm.outputs.cache-hit && startsWith(inputs.rake-job, 'almalinux-9') }}
+        run: |
+          if find fluent-package/yum/repositories -type f -name 'fluent-package-*.rpm' \
+            ! -name '*debug*' ! -name '*.src.rpm' | xargs rpm -qp --requires | grep -qE "GLIBC_2\.(3[6-9]|[4-9][0-9])|OPENSSL_3\.[1-9]"; then
+            echo "::error::Incompatible dependencies (GLIBC_2.36 or later, OPENSSL_3.1.x or later) were detected!"
+            exit 1
+          else
+            echo "Compatibility with GLIBC_ and OPENSSL_ ABI check passed. No incompatible dependencies found."
+          fi
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         if: ${{ ! steps.cache-rpm.outputs.cache-hit && inputs.next-major }}
         with:
--- a/fluent-package/yum/almalinux-9-aarch64/from
+++ b/fluent-package/yum/almalinux-9-aarch64/from
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-arm64v8/almalinux:9
+arm64v8/almalinux:9.2
--- a/fluent-package/yum/almalinux-9/Dockerfile
+++ b/fluent-package/yum/almalinux-9/Dockerfile
@@ -15,10 +15,11 @@
 # specific language governing permissions and limitations
 # under the License.
 
-ARG FROM=almalinux:9
+ARG FROM=almalinux:9.2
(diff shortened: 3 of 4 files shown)


kenhys:
How to verify it:

* build it with the following command:
  * DEBUG=yes rake yum:build YUM_TARGETS="almalinux-9"
* encure reuired symbols
  * rpm -q --requires ./fluent-package-6.0.0-1.el9.x86_64.rpm  |grep -E "OPENSSL|GLIBC"
  * No GLIBC_2.35 and OPENSSL_3.4.0.

kenhys:
Now it will be checked like this:

image

.github/workflows/_build_rpm_reusable.yml
```diff
@@ -42,6 +42,16 @@ jobs:
         if: ${{ ! steps.cache-rpm.outputs.cache-hit }}
         run: |
           rake yum:build YUM_TARGETS=${{ inputs.rake-job }}
+      - name: Ensure RHEL9 compatibility
+        if: ${{ ! steps.cache-rpm.outputs.cache-hit && inputs.rake-job == 'almalinux-9' }}
```
Watson1978:
```suggestion
        if: ${{ ! steps.cache-rpm.outputs.cache-hit && startsWith(inputs.rake-job, 'almalinux-9') }}
```
Please contains `almalinux-9-aarch64`

kenhys: Thanks!

.github/workflows/_build_rpm_reusable.yml
```diff
@@ -42,6 +42,16 @@ jobs:
         if: ${{ ! steps.cache-rpm.outputs.cache-hit }}
         run: |
           rake yum:build YUM_TARGETS=${{ inputs.rake-job }}
+      - name: Ensure RHEL9 compatibility
+        if: ${{ ! steps.cache-rpm.outputs.cache-hit && startsWith(inputs.rake-job, 'almalinux-9') }}
+        run: |
+          if find fluent-package/yum/repositories -type f -name 'fluent-package-*.rpm' \
+            ! -name '*debug*' ! -name '*.src.rpm' | xargs rpm -qp --requires | grep -qE "GLIBC_2\.35|OPENSSL_3\.4\.0"; then
```
Watson1978: The current `grep -qE "GLIBC_2\.35|OPENSSL_3\.4\.0"` only matches the two specific symbols we happened to hit this time. The guard's real goal is to reject any requirement that goes *above* the el9 baseline — i.e. above what the lowest EL9 minor we support actually provides (glibc up to 2.35, OpenSSL 3.0.x). If a future 9.x introduces e.g. `GLIBC_2.36` or `OPENSSL_3.5`, the current regex would let it slip through. Matching a range catches the next drift automatically:

```
GLIBC_2\.(3[6-9]|[4-9][0-9])|OPENSSL_3\.[1-9]
```

Note that `GLIBC_2.35` itself should stay allowed: as the logs below show, every EL9 (AlmaLinux 9.0 and Oracle Linux 9 alike) already provides `GLIBC_2.35`, so it may even be enough to check OpenSSL alone (treat `OPENSSL_3.1.0` or later as an error)

```
$ docker run --rm almalinux:9.0 bash -c '
  rpm -q --provides glibc        | grep -oE "GLIBC_[0-9.]+"   | sort -uV | tail -3
  rpm -q --provides openssl-libs | grep -oE "OPENSSL_[0-9.]+" | sort -uV | tail -3
'
GLIBC_2.33
GLIBC_2.34
GLIBC_2.35
OPENSSL_3.0.0
OPENSSL_3.0.1
```
```
$ docker run --rm oraclelinux:9 bash -c '
  rpm -q --provides glibc        | grep -oE "GLIBC_[0-9.]+"   | sort -uV | tail -3
  rpm -q --provides openssl-libs | grep -oE "OPENSSL_[0-9.]+" | sort -uV | tail -3
'
GLIBC_2.33
GLIBC_2.34
GLIBC_2.35
OPENSSL_3.3.0
OPENSSL_3.4.0
OPENSSL_3.5.0
```

Watson1978 (approved):
👍🏻

PR #1088, opened 2026-07-10
yum: keep compatibility with older RHEL 10.x

kenhys:
yum: keep compatibility with older RHEL 10.x

When building with latest RHEL 10.x, it breaks ABI compatibility with
older RHEL 10.x because it requires GLIBC_2.40.

This is because glibc 2.40 symbols were backported into its glibc 2.39.
It doesn't exist on RHEL 10.1 and earlier version.

Thus build environment should be pinned to 10.0 for keeping
compatibility with all range of RHEL 10.x.


kenhys:
rebase with #1087

kenhys:
GLIBC_2.4 should be ignored.

Watson1978 (approved):
👍🏻

PR #1089, opened 2026-07-10
Backport(v6) yum: keep compatibility with older RHEL 9.x (#1087)

kenhys:
When building with latest RHEL 9.x, it breaks ABI compatibility with older RHEL 9.x because it requires GLIBC_2.35 and OPENSSL_3.4.0.

This is because glibc 2.35 symbols were backported into its glibc 2.34. It doesn't exist on RHEL 9.6 and earlier version.

Thus build environment should be pinned to 9.2 for keeping compatibility with all range of RHEL 9.x.

See "Werk #18941: Fix installation failure on AlmaLinux 9 due to GLIBC_2.35 symbol mismatch"
https://checkmk.com/werk/18941


Watson1978 (approved):
👍🏻

PR #1093, opened 2026-07-14
ci: use composite action instead of reusable workflow

kenhys:
It makes easier to understand which is called from parent workflows.


.github/actions/test-on-docker/action.yml
```diff
@@ -0,0 +1,65 @@
...
+    description: 'Additional docker options'
+    required: false
+    default: ""
+  centos-stream:
+    description: 'CentOS Stream version if applicable'
+    required: false
+    default: ""
+  test-label:
+    description: 'Label for the test step'
+    required: true
```
Watson1978: [nits] new workflow does not use `test-label`.
We can remove this.

Watson1978 (approved):
👍🏻

.github/actions/test-on-docker/action.yml
kenhys: Thanks, removed.

PR #1094, opened 2026-07-14
Backport(v6) ci: use composite action instead of reusable workflow (#1093)

kenhys:
It makes easier to understand which is called from parent workflows.


Watson1978 (approved):
👍🏻

PR #1097, opened 2026-07-24
Bump master version

kenhys:
6.0.x will not released from master branch,
    so bump internal version to 6.1.0.


Watson1978 (approved):
👍🏻

PR #1103, opened 2026-08-06
Skip building the current major version in the next major lane

Watson1978:
"Build next major" ran "rake yum:build" twice: once for the current version and once more for the v7 one in v7-test. The first build's artifacts are discarded, since all three upload-artifact steps for them are guarded by next-major != 'true', so it was pure waste. The deb counterpart already guards the step this way; only the rpm one was left behind.

Measured on the AlmaLinux 10 x86_64 lane of run 30994758570:

  * Build             rake yum:build (v6)   8m27s
  * Build next major  rake yum:build (v6)   6m56s  <- discarded
                      rake yum:build (v7)   6m16s

To keep the RHEL9 and RHEL10 ABI checks meaningful, move them after the v7 build and point them at whichever repositories directory the lane actually produced. This makes the next major lane check the v7 packages instead of repeating what the "Build" lane already checks.

Also fail loudly when no rpm is found to check, so that a wrong directory cannot be mistaken for a clean result.

### DIFF
--- a/.github/actions/build-rpm/action.yml
+++ b/.github/actions/build-rpm/action.yml
@@ -37,36 +37,10 @@ runs:
         gem install serverspec --no-document
 
     - name: Build rpm with Docker
-      if: ${{ steps.cache-rpm.outputs.cache-hit != 'true' }}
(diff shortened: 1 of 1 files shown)


kenhys (approved):
LGTM.

